PR #4, opened 2016-08-18
2014stable

natedill:
Added ability to specify non-zero periodic flux automatically from information in the asgs_config so the user doesn't have to edit the fort.15 template to get the proper river boundary conditions for models that use steady-state river inputs(e.g. Louisiana models with the Mississippi and Atachafalaya rivers)


jasonfleming:
@natedill somehow I did not realize this pull request was available. Is there any reason I should hold off on accepting this pull request? If not, I will go ahead and accept it.

natedill:
Hi Jason,

I don't see any reason not to accept it.  The biggest change was adding a
feature that allows the operator to specify constant periodic flux in the
config script.  This required adding some variables and file checking to
asgs_main.sh, some changes to control_file_gen.pl to handle filling in the
period flux data in a fort.15 template, and the addition of a script to
calculate the flux.  I also added an example config script and fort.15
template.

The changes shouldn't do anything unless the operator configures their
config script to use it, and provides fort.15 template with %PEROIDICFLUX%
line in it in place of the normal periodic flux boundary condition data.

The other additions I think are all different utilities.

Cheers,
Nate

PR #74, opened 2019-05-14
Issue 31

wwlwpd:


jasonfleming:
Hey this looks good so far. I'd like to use the fixed version of NAMtoOWIRamp.pl for the exercise on 28 May if possible, can we change the PR so it merges to 2014stable branch instead of master? I periodically merge 2014stable to master. Also, I'd like this PR to remove the older NAMtoOWI.pl script that we don't actually use any more to avoid confusion. Finally, I think the issue of memory usage is mainly the result of pushing all the data from the full range of files into @ugrd @vgrd and @atmp roundabout lines 695-714 in &getGrib2 ... this is ok for a small set of files but for e.g. 2 weeks of data (esp. for higher resolution) it seems to fill up memory ... these arrays then copy their data into another set of arrays of the same size in &toOWIformat, doubling the memory use ... can you have a look at the code to see if I am reading this right?

wwlwpd:
@jasonfleming - I retargeted this PR to 2014stable, will work on your other requests in additional commits

wwlwpd:
>I think the issue of memory usage is mainly the result of pushing all the data from the full range of 
>files into @ugrd @vgrd and @ATMP roundabout lines 695-714 in &getGrib2 ... this is ok for a small 
>set of files but for e.g. 2 weeks of data (esp. for higher resolution) it seems to fill up memory ... these 
>arrays then copy their data into another set of arrays of the same size in &toOWIformat, doubling the 
>memory use ... can you have a look at the code to see if I am reading this right?

I believe you are correct. Without a complete refactor, we have 2 options:

1. (not recommended) push scalar references to the original array elements into the new arrays, this might shave a few bytes off of each entry because it needs the size of 1 memory reference versus the the size of the number in the array slot. I don't think this would make much of a difference and scalar references are greatly discouraged in Perl best practices.

2. drain the source arrays by shifting them off the old and unshifting them into the new; to maintain the integrity of the array being iterated over, it'd have to be done in reverse. This s destructive on the source arrays, so if they're used elsewhere then that would complicate this approach.

3. a less efficient version of #2 would be to just delete the source arrays after the existing loops are in place; this would alleviate concerns of corrections for the new arrays, but if the old arrays are needed after this processing, then that would cause the same complications as #2.

If we want to refactor and if there is limited dependency on previous elements then it might be possible to rewrite the whole script to do the conversion inside of the loops that are reading the files - this would provide the maximum amount of memory efficiency, but it is also probably a total rewrite.

I am going to make a proposed change using option #2. I would be great to have a set of test data so that I can ensure that the resulting file(s) are identical both before and after the change.

wwlwpd:
Cancelling in favor of #79. Comments above are still valuable.

PR #76, opened 2019-05-17
added instructions for locally installer Perl

wwlwpd:
perlbrew can be used to create a fully controlled perl environment
that may persist between logins. Manual nstructions for doing this is
added in this commit.

Note: I hope to automate this, but these instructions are pretty straight forward and only need to be done once.


wwlwpd:
superseded by #77

PR #85, opened 2019-05-30
Replaced all shebangs with '/usr/bin/env perl'

wwlwpd:
Issue 83: First commit to get repo perlbrew aware.


jasonfleming:
Excellent, I had no idea we had so many perl scripts. :-)

PR #108, opened 2019-07-25
Default base directory of PREPPEDARCHIVE and HINDCASTARCHIVE to SCRATCH

wwlwpd:
Issue 101: Added a new variable option for the config file to set the base
directory for PREPPEDARCHIVE and HINDCASTARCHIVE. If RUNARCHIVEBASE is not
otherwise set in the configuration file, SCRATCHDIR is used. SCRATCHDIR also
now defaults to $SCRATCH (common envar on HPC systems) unless SCRATCHDIR is
explicitly defined in the asgs_main.sh configuration file (typically defined).

### DIFF
--- a/asgs_main.sh
+++ b/asgs_main.sh
@@ -45,6 +45,14 @@ echoHelp()
   echo "-h : show help"
   exit;
 }
+
+# reads/rereads+rebuilds derived variables
+readConfig()
+{
+  . ${CONFIG}
+  RUNARCHIVEBASE=$SCRATCHDIR
+}
+
 #
 # subroutine to check for the existence of required files that have
 # been specified in config.sh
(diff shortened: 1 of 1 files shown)


wwlwpd:
I did not test these changes in a full run, but I did test that the general concept of the change works fine. That concept is that the "include" (. $CONFIG) is encapsulated inside of a bash function. That function call (readConfg) was substituted for all instances of ". $CONFIG". The one spot I did not show to work in this was was around line 1769,

>trap 'echo Received SIGUSR1. Re-reading ASGS configuration file. ; readConfig' USR1

There are other opportunities to encapsulate the inclusion of the other files there functions are defined, but I didn't do it because I wanted this to be scoped to the original charge.

wwlwpd:
And per milestone instructions, this PR is targeting the master branch.

Issue #133, opened 2019-09-02
asgs-email.pl should accept space delimited email addresses like sendmail


wwlwpd:
It is a bug that it can't do this since it's supposed to be a drop in replacement for sendmail.

wwlwpd:
PR #134 addresses this.

wwlwpd:
merged, closing

PR #134, opened 2019-09-02
--to accepts comma or space delimited email address lists now

wwlwpd:
Issue 133: This change allows the --to option to utilize a list of
email addresses that are delimited by either spaces or commas. Since
this script is meant to be mostly a drop in replacement for sendmail,
it and sendmail delimits email addresses by spaces, it's important
to support this.


wwlwpd:
reported in issue #133

Issue #144, opened 2019-09-08
provide an "asgs-shell" wrapper for testing, debugging, and probably operational use


wwlwpd:
This is just a shell script that is going to be placed into $install_path/bin. It's purpose is to begin an interactive bash shell with the asgs-brew.pl environment explicitly set. This can be used for testing, debugging, and likely operations.

Issue #140, opened 2019-09-06
Deprecate Date::Pcalc in favor of Date::Calc


wwlwpd:
https://metacpan.org/pod/Date::Calc
https://metacpan.org/pod/distribution/Date-Pcalc/Pcalc.pod

Date::Calc is the C/compiled version of Date::Pcalc; the latter is the "pure Perl" version. Date::Calc is by the same author as Date::Pcalc, has the same set of methods, and installs cleanly and easily - as opposed to Date::Pcalc. 

It should be a drop-in replacement.

wwlwpd:
Historical note, Date::Pcalc was originally chosen because the required Perl modules were distributed with the application code. Since we are moving to perlbrew and away from distributing non-ASGS specific Perl modules, it no longer makes sense to use the "pure Perl" version of the module. Date::Pcalc also must be installed interactively and with no tests since it requires one to choose the between the [c]ompiled version or [p]ure perl version during installation. Some tests also fail for Date::Pcalc. Both issues require us to treat Date::Pcalc differently from all other modules installed into the perlbrew environment.

Issue #136, opened 2019-09-04
add python set up as a step


wwlwpd:
this issue turned into: install python 2.7.17 locally (similar to how we handle perl)

wwlwpd:
supercedes #159 since it was the the half baked python environment on LSU resources (qb, supermic) that was causing this step to break only over there.

Issue #135, opened 2019-09-04
perlbrew invoked via asgs-brew.pl is not making Perl modules available


wwlwpd:
noticed by @wwlwpd and @jasonfleming

wwlwpd:
this should include as a deliverable a test Perl script (Test::More) that uses use_ok to verify that all modules are available; failure of the script should cause the postcondition check to fail

PR #146, opened 2019-09-18
Major update to asgs-brew.pl, major addition to interactive environment.

wwlwpd:
Issue #157: Patch that was being applied is now in perlbrew via
Devel::PatchPerl, so there is no need to apply ourselves - in fact,
it causes a non fatal warning which causes confusion. (tested on stampede2
which is why it was needed in the first place).

Issue #144: Created a shell wrapper that makes available the ASGS that
asgs-brew.pl makes available. Added some example "helper" to show what is
possible in the future.  Moved INSTALL to cloud/general, removed mention of
local::lib in the manual perlbrew installation instructions.

Issue #140: Removed installation of Date::Pcalc because it required
being treated differently than the other modules and its tests were
failing. Replaced all instances of Date::Pcalc in scripts with
Date::Calc. This commit does NOT remove the old version(s) of Date::Pcalc
that are distributed with ASGS, but that (and removing all CPAN
modules that are distributed) is the next step since the installation
of all non-custom modules is being handled using perlbrew and cpanm.

Issue #135: Adding better checks for failure, removed use of local::lib,
which seems to have been causing some confusion.

Issue #139: Added netcdf/hdf5 init scripts missing from a previous commit.

Issue #127: Also added a "setup-env" step and partitined out
the setting of specific environmental variables to the appropriate
step. Added a MACHINENAME variable that reflects the value of
the required --machinename flag..

Added step to install required Python libraries.

Issue #136: Covers pika, netCDF5, and the pptx modules. Did not verify
that RabbitMQ support is provided.

Issue #139: Split steps, NetCDF4 step is still dependant on HDF5; so
to enforce the precondition on the netcdf4 step is identical to the
postcondition check for hdf5.

Issue #138: Added options to OpenMPI's configure command. This reduced
build time significantly. Mostly it was the tracing libraries and tools
that were disabled.

Addition items not tied to a specific issue...

asgs-init.sh now asks a series of questions to create the asgs-brew.sh
command used because there was a need to source platforms.sh before,
therefore the need to properly bootstrap the environment was required
even for the initial running of asgs-brew.pl; this bootstrap script
can be distributed as a single point of installation for most systems
that are supported in platforms.sh; that is it can be distributed
alone and when run, checkouts the ASGS repo if not found - thus providing
access to all required files needed for set up

updated documentation

A refining, imported properties.sh for future access to run.properties
settings

Added sourcing steps for logging.sh and platforms.sh to asgsh generation
step.

Updated error messages to reflect supported subcommands; on start
the shell now imports the functions defined in platform.sh and
logging.sh; then it attempts to run 'env_dispatch' based on the
value of $MACHINENAME as set by asgs-brew.pl. This is still
experimental and no practical use has been devised. However, these
are essential to asge_main.sh, so any issues discovered by doing
this should be fixed in the shell rather than being avoided.

reload will reread the profile for a current session, thus re-parsing
the config file and looking for state file, log, and properties info.
save now works with no NAME param if session is currently loaded, calls
the reload routine if used with a currently loaded session; useful for
when one is first setting up a config or operations for a new storm

Using INSTANCENAME that can be grep'd out of the config file, we
can discover the location of the state file in the scratch dir; from
there we can discover the RUNDIR, SYSLOG, and location of the primary
run.properties file. Using this information, implemented 'watchlog'
that 'tail -f's the log file. No derived information is save in the
named profile since it should be discovered by first inspecting the
configuration file.

updated Vagrant file

added load/save/delete

make sure that all include Perl and Python scripts can pass
their respective compiliation phases; a byproduct is a set of
.pyc files, which are removed - but these could prove to be useful
at a later date. Update existing Perl scripts so they pass minimal
checks.


wwlwpd:
@jasonfleming we should discuss this before merging it.

jasonfleming:
Hey @wwlwpd I just tried the asgs-brew.pl from master and found that when the resulting ASGS started, it ended up having an issue with not being able to find Pcalc.pm ... I am going to give this a try now. And if it works, I am going to merge it (unless you recommend that I don't). One thing we may want to think about is having multiple versions of ASGS installed without colliding (if that is not in here already) as well as having an uninstall feature. FYI @akheir

wwlwpd:
Runing asgs-brew.pl in master right now will likely result in issues, I've been on this branch for a while now so I don't know how it will actually behave. There is currently a conflict with master, I will fix this and push an updated PR.

Some notes:

1. **One minor incompatibility with the current setup** - Date::Calc is now required over Date::Pcalc, so the fix should be to just install Date::Calc and make it available via PERL5LIB

2. Let's talk about multiple versions living side by side so I better understand the requirements

3. LSU resources are presenting an issue due to their Python environments and head node process time limits. This will have to be part of my next focus.

wwlwpd:
Sorry @jasonfleming I didn't see that you approved this. I will do the merge.

wwlwpd:
And thank you!

Issue #162, opened 2020-01-07
update platforms.sh to be consistent with adcirc's work/cmplrflags.mk for lonestar


wwlwpd:
This just entails replace lonestar with lonestar5 and using the proper hostname (ls5.tacc.utexas.edu)

PR #166, opened 2020-01-09
asgs-brew changes needed for the 2019stable release.

wwlwpd:
Issue #136: Added step to install python 2.7.17 to ensure
that the python environment is consistent among the various
platforms that must be supported. This was necessary because
a Perl module that ASGS requires was not able to get built
properly on some platforms due to the provide python providing
an incorrect openssl environment. Beyond that, it ensures more
stability.

Issue #161: Updated documentation and added an INSTALLING file
in the root ASGS directory describing how to quickly install
ASGS on a new platform.

Issue #162: Updated platforms.sh to be more consistent with
ADCIRC's cmplrflags.mk file and the specific platform environments
contained therein.


wwlwpd:
Thank you.

How is `lonestar5.tacc.utexas.edu` used? It doesn't resolve to an IP address whereas ls5 does:

    [@smic1 asgs]$ ping lonestar5.tacc.utexas.edu
    ping: unknown host lonestar5.tacc.utexas.edu
    [@smic1 asgs]$ ping ls5.tacc.utexas.edu
    PING ls5.tacc.utexas.edu (129.114.62.196) 56(84) bytes of data.
    64 bytes from login1.ls5.tacc.utexas.edu (129.114.62.196): icmp_seq=1 ttl=52 time=10.3 ms
    64 bytes from login1.ls5.tacc.utexas.edu (129.114.62.196): icmp_seq=2 ttl=52 time=10.3 ms
    64 bytes from login1.ls5.tacc.utexas.edu (129.114.62.196): icmp_seq=3 ttl=52 time=10.3 ms

jasonfleming:
Its just a name for the opendap path where the results are posted, and for the run.properties file etc ... supermic is actually smic.hpc.lsu.edu etc ... HPCENV is just a more human recognizable name for the machine but I agree that it is confusing b/c it looks like a FQDN

Issue #169, opened 2020-01-09
allow asgsh to support & manage full environments


wwlwpd:
See #167 for context

Issue #167, opened 2020-01-09
multiple brewed ASGSes (or none) in a single user account on HPC


jasonfleming:
We often need to have multiple ASGSes installed on a single user account on an HPC system (e.g., 2019stable for production alongside master or a feature branch). Does asgs-brew accommodate that? Related to that, an HPC user may want to completely uninstall the ASGS, or may have other activities going on the HPC that should not use the custom python and/or perl that the ASGS environment provides, is it possible uninstall the ASGS or temporarily disable the asgs-specific environment?

jasonfleming:
Also, asgs-brew is modifying the ~/.bash_profile ...   do different HPC machines require modifying different startup files like ~/bashrc or ~/.profile?

jasonfleming:
How do the modules systems on these platforms interact with the environment variables set by asgs-brew? For example, if an Operator has certain modules loaded when they install, do the environment variables set by asgs-brew incorporate those PATHs, so they can't be circumvented thereafter by unloading the module?

jasonfleming:
Does ~/.bash_profile get appended by asgs-brew or overwritten or does asgs-brew go through and rewrite the env variable settings? In other words, how to prevent collisions between asgs-brew settings and other unrelated env variable settings set by the Operator in that file?

wwlwpd:
`~/.bash_profile` only gets updated after perlbrew is run, but I was going to remove that actually since we should direct everyone through asgsh. I'll add an issue to do that. 

Other than that, the only thing really preventing this is that `asgsh` uses a statically defined environment file. Allowing full environments side by side through a single asgsh would be a great feature.  I'll add this as another issue.

To do it manually, you'd have to rename `$HOME/.asgs-brew.sh` to not clobber, then copy `$install_path/bin/asgsh` to something like `$install_path/bin/asgshX`. And even then you'd have to avoid save profile names across all installations. So definitely a pain, but this can be added without too much work.

wwlwpd:
Created #168 and #169. @jasonfleming is either or are both blockers for stable2019 or would you like to consider them either bug fixes or target master? I can prioritize them and it wouldn't be much work to get this working.

wwlwpd:
Oh and `~/.bash_profile` is intelligently updated or at least there is an attempt to be smart about it.

jasonfleming:
Ok thank you, I am still going through the installation work and would like to get comfortable with it for 2019stable before we make it official. So yes I think it is right to continue to focus on it in the immediate term.

[...]

PR #175, opened 2020-01-17
Made python environment more consistent with perl environment.

wwlwpd:
Issue #167: Modified python to install python 2.7.17 into $HOME
much like perlbrew does for perl. The directory structure used
by perlbrew is reflected in the path now used to install python.
This means that for multiple ASGS installations (managed by asgsh),
they will all use the same python like they will currently use the
same perl.

Fixed some isses with a couple of existing Perl scripts not finding
custom modules located in the ./PERL directory. Moved modules that
are installed via cpanm and the perl-modules step of asgs-brew.pl
into ./PERL/deprecated. The intent is to remove them very soon and
make ./PERL only for custom Perl libraries (will likely rename this
directory also).  Install Perl::Tidy module and installs our "standard"
perltidyrc file as $HOME/.perltidyrc so that we can assume a basic
level development environment workflow for Perl scripts that includes
tidying the files.

Also covers issues: #169 (handle multiple asgs environments),


jasonfleming:
I agree that renaming or eliminating ./PERL in favor of our custom perl environment is a great idea. Can we complete that process before releasing 2019stable?

Issue #168, opened 2020-01-09
asgs-brew shouldn't modify ~/.bash_profile


wwlwpd:

wwlwpd:
will be taken care of in #167 / #169

PR #176, opened 2020-01-17
Made python environment more consistent with perl environment.

wwlwpd:
Issue #167: Modified python to install python 2.7.17 into $HOME
much like perlbrew does for perl. The directory structure used
by perlbrew is reflected in the path now used to install python.
This means that for multiple ASGS installations (managed by asgsh),
they will all use the same python like they will currently use the
same perl.

Fixed some isses with a couple of existing Perl scripts not finding
custom modules located in the ./PERL directory. Moved modules that
are installed via cpanm and the perl-modules step of asgs-brew.pl
into ./PERL/deprecated. The intent is to remove them very soon and
make ./PERL only for custom Perl libraries (will likely rename this
directory also).  Install Perl::Tidy module and installs our "standard"
perltidyrc file as $HOME/.perltidyrc so that we can assume a basic
level development environment workflow for Perl scripts that includes
tidying the files.

Also covers issues: #169 (handle multiple asgs environments),
and #168 (perlbrew step shouldn't modify $HOME/.bash_profile).


wwlwpd:
>Can you create an issue for this?
Created issue #183 .

>have asgs-brew clone and setup Nate's perl modules because they do a lot of nice adcirc-specific things
Created issue #184

wwlwpd:
>Some of these modules and perl scripts are from Nate Dill's OurPerl git repo (AdcGrid.pm, tpxoBoundaryInterp.pl, riverFlow.pl) and are not used in an automated way, so we can probably not worry too much about them for now.

I like the idea of carrying such modules along or getting them directly from GH. Ideally, it's be great to package all required ASGS perl modules up on CPAN even if it is to make it available for us via cpanm.

wwlwpd:
@jasonfleming requested change has been made.

PR #179, opened 2020-01-18
fixed issue 178, removed duplicated init_frontera() block in platforms.sh

jasonfleming:
Also removed QOS=vip setting b/c not used on Frontera.


wwlwpd:
Nice catch!

wwlwpd:
Generally if the code gets approved, the reviewer should merge it. There are ways to improve this workflow, but since it's the 2 of us right now I think it's okay. So I approve it and I am going to merge it. You would do the same for me.

PR #198, opened 2020-02-17
Tga pull station fixes

tgasher:
Fixed a bug in pullStationTimeSeries.f90 where it had an erroneous hard-coded file ID.  Also modified adcmesh.f90 and pullStationTimeSeries.f90 so that, if a requested station output's coordinates exactly match a node's coordinates, then the nodal values are output directly.

### DIFF
--- a/output/adcmesh.f90
+++ b/output/adcmesh.f90
@@ -1956,6 +1956,7 @@ real(8) :: x1, x2, x3 ! longitude temporary variables
 real(8) :: y1, y2, y3 ! latitude temporary variables
 real(8) :: subArea1, subArea2, subArea3, TotalArea
 integer :: e
+logical :: exactmatch !whether station output is exactly at a node
 
 if (station%elementFound.eqv..false.) then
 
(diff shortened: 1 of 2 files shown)


tgasher:
A note for future usage:  I just tried this on an alternate test case and found there to be disagreement in the lon/lat coordinates in an .nc data file as compared to what was in the original mesh.  My guess is that the netCDF and Fortran libraries don't quite follow the same conventions during some aspect of the reading/writing process.  Ultimately, this means that to ensure a good match, one may need to require that the fort.14 file is used determine node coordinates, rather than relying on what's in the netCDF file.  I have verified that hard-wiring pullStationTimeSeries.f90 to use the fort.14 file for nodal coordinates alleviated this issue across a test of 121 stations, of which 33 failed otherwise.  

I spent some time looking into where the loss of precision occurs.  I didn't get to a final answer, but here's what I found.  First, an example of non-matching coordinates:  
Add this code to adcmesh.f90 at the end of this IF statement `if(station%elementFound.eqv..true.)` to output lon, lat, weight info:  
`write(*,'(f,f,l)')station%lon,station%lat,exactmatch
write(*,'(3f)')m%xyd(1,m%nm(station%elementIndex,1)),m%xyd(2,m%nm(station%elementIndex,1)),station%w(1)
write(*,'(3f)')m%xyd(1,m%nm(station%elementIndex,2)),m%xyd(2,m%nm(station%elementIndex,2)),station%w(2)
write(*,'(3f)')m%xyd(1,m%nm(station%elementIndex,3)),m%xyd(2,m%nm(station%elementIndex,3)),station%w(3)`
produces this for one sample point whose coordinates exactly match what's in the fort.14 file:  
     -80.6785508475000057      28.3587202078999994 F
     -80.6790977842000103      28.3590750002000007       0.0000000000000000
     -80.6785508475000199      28.3587202078999994       0.0000000000000000
     -80.6786178196000066      28.3592745706000002       0.0000000000000000

The node of interest corresponds to this line in my fort.14 file:  
   4021886  -80.6785508475   28.3587202079   -0.0342739618

Looking at IEEE double precision numbers, 
the fort.14 file's value:  -80.6785508475 
is equal to 
the real(8) interpretation of this value in the code:  -80.6785508475000057
but neither are equal to 
the real(8) value stored in the .nc file:  -80.6785508475000199
This suggests the .nc file's coordinates are the ones that're suffering a degradation in precision somewhere along the way.

tgasher:
@jasonfleming right, that's exactly what I did.  I actually started by testing equality because I wanted to test how the precision came out.  And, yes, a node-based data extraction option would be very welcome.

jasonfleming:
Hey Taylor, I've now resolved this with PR #325 for issue #318 ... you can now specify a list of nodes in a file (analogous to existing ability to specify a stations file) for pullStationTimeSeries.f90 and it will get the values at that list of nodes, without spatially interpolating, or trying to "find" them. :-)

Issue #165, opened 2020-01-09
create docker base image (no adcirc)


wwlwpd:
Install everything except adcirc itself

wwlwpd:
Docker image (instructions on getting it set up is at the link):

https://hub.docker.com/r/asgsdockerhub/2019stable

A pull request is pending.

Dockerfile used to build the image (not yet committed to master and not part of the 2019stable release):

https://github.com/wwlwpd/asgs/blob/issue-165/Dockerfile

One thing to note is that gfortran is the Fortrancompiler available on the docker image and is used for compiling ADCIRC via "initadcirc" in the ASGSH shell.

PR #204, opened 2020-03-29
Adding initial Dockerfile

wwlwpd:
Issue #165: Adding a Dockerfile to support the evaluation,
exploration, and local use of the 2019stable release of ASGS.
The Dockerfile is used by a developer or maintainer when after
a release when they wish to make a Docker image of the environment
available at https://hub.docker.com.

This change also includes a minor change to the script that installs
the Perl modules because during the development of the Dockerfile,
it was discovered that the upstream version of Date::Handler had some
failing tests that were preventing the module from being installed.
This was through no fault of ASGS, nor does it appear that this test
failing affects ASGS in any negative way.


wwlwpd:

PR #210, opened 2020-04-18
Allow advisoryMonitor.sh to fallback to http (non-ssl) for testing.

wwlwpd:
Issue 209: Added code to fallback to http (non-ssl) when there is
a 599 HTTP response code, indicating in this case that the connection
to port 443 (https) was refused. Cleaned up a few things while in
there; the first 4 commandline arguments are preserved, but added
a few more to allow for specifying FTP hostname/ip, RSS hostname/ip,
trigger type, etc. Any other cleaning up preserved the existing
functionality.

This was mainly done to make it easier to test ad hoc FTP and WWW
servers that might be set up to mimic the NHC infrastructure, thus
providing another and more realistic way to do integration testing
(and real investigation/research).

### DIFF
--- a/get_atcf.pl
+++ b/get_atcf.pl
@@ -26,15 +26,14 @@
 # You should have received a copy of the GNU General Public License
 # along with the ASGS.  If not, see <http://www.gnu.org/licenses/>.
 #
-$^W++;
+use warnings;
 use strict;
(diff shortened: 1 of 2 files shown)


wwlwpd:
redirecting into integration branch for now

Issue #215, opened 2020-04-18
Create initial "replay script" for use with NHC-looking WWW and FTP servers used by ASGS


wwlwpd:
Include instructions on set up and how to set up with the WWW and FTP docker files

PR #219, opened 2020-04-20
Added inital tool for easily replaying storms.

wwlwpd:
Issue #215: This is the initial script for replaying storms
in a configurable way for testing, development, training,
conducting readiness drills. There are lots of opportunities
to determine how to extend this script. It is meant to run on
a server that is configured with FTP and RSS (via http) which
mimicks how the NHC provides advisories during actual events.
Documentation for the script is provided via POD and can be
viewed using "perldoc ./replay-storm.pl".

In this first version, index-at.xml files are updated for each
storm by adding an entry for each storm. Storms are updated in
place, but the existing scheme allows each storm among many to
exist the XML exactly one time.

### DIFF
--- a/input/sample_advisories/ftp_rotate.sh
+++ b/input/sample_advisories/ftp_rotate.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+STORM=05
+NAME=DORIAN
+YEAR=2019
+START=01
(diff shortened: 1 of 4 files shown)


wwlwpd:

Issue #184, opened 2020-01-18
have asgs-brew clone and setup Nate's perl modules because they do a lot of nice adcirc-specific things.


wwlwpd:

wwlwpd:
mentioned in #176

wwlwpd:
@jasonfleming, I can confirm that in the current ASGS Shell environment:

asgs/PERL is included in the `PERL5LIB` environment, which makes `AdcGrid.pm` available. This can be confirmed in `asgsh` by running the command `perl -MAdcGrid` (hit `ctrl-c` to exit).

Also, the following scripts have all requirements fulfilled and pass the `perl -c` check (these are included in the `verify` check):

* tpxoBoundaryInterp.pl
* riverFlow.pl

So I can confirm that these scripts are "supported". Did you wish for them to be made available in some other way at this time? If not we can close this issue, but I am also happy to implement some further concept of "supported" that you'd like to cover this issue. Thanks!

wwlwpd:
BTW, here is the full output of `verify` which checks all Perl modules, scripts, and Python scripts:

`
verifying Perl Environment:
/home/estrabd/perl5/perlbrew/perls/perl-5.28.2/bin/perl
ok 1 - require IO::Socket::SSL;
ok 2 - require Config::Tiny;
ok 3 - require Date::Calc;
ok 4 - require Date::Handler;
ok 5 - require DateTime;
ok 6 - require DateTime::Format::Builder;
ok 7 - require Email::Sender::Simple;
ok 8 - require Email::Sender::Transport::SMTP::TLS;
ok 9 - require Email::Simple::Creator;
ok 10 - require HTTP::Tiny;
ok 11 - require List::Util;
ok 12 - require Math::Trig;
ok 13 - require Net::FTP;
ok 14 - require Params::Validate;
ok 15 - require Test::More;
ok 16 - require Time::Local;
ok 17 - require Try::Tiny;
ok 18 - require Perl::Tidy;
1..18
verifying Perl scripts can pass compile phase (perl -c)
ok ./nhc_advisory_bot.pl
ok ./storm_track_gen.pl
ok ./NAMtoOWIRamp.pl
ok ./get_atcf.pl
ok ./input/sample_advisories/advisory_xml_filler.pl
ok ./input/ptFile_gen.pl
ok ./control_file_gen.pl
ok ./stage_discharge.pl
ok ./set_flux.pl
ok ./tides/tpxoBoundaryInterp.pl
[...]

PR #220, opened 2020-04-20
Added AdcGrid to the Perl module check available in asgsh.

wwlwpd:
Issue #184: Access to AdcGrid is implicitly proved during the
`perl -c` check, but adding it to the Perl module verification
check so there is no doubt.


wwlwpd:

PR #234, opened 2020-05-15
Run ready 2020 rebased

wwlwpd:
@jasonfleming - currently doing a quick QA over it. I did a full rebase over the current master and it was a little tricky. Will ping you when it's all clear to merge from my end. Don't let that stop you from looking over the changes, though. Biggest "risky" change is the update to `./get_atcf.pl`.


wwlwpd:
Found a another default value issue with ./init-asgs.sh for where to install things. Fix is contained here. Still looking.

wwlwpd:
Missing XML::RSS, will be fixing shortly

Issue #239, opened 2020-05-19
qscript.pl can't handle if any *WALLTIME doesn't have a 2-digit (or zero-padded) hour


wwlwpd:
Should be able to handle D:DD:DD rather than 0D:DD:DD (zero padded). Fix forthcoming.

PR #240, opened 2020-05-19
Modified qscript.pl to handle single digit hour *WALLTIMEs.

wwlwpd:
Issue #239: Updated regex that extracts hour:minutes:seconds from
the *WALLTIME passed via --walltime option so that it can handle
single digit hours (and minutes and seconds).

Now accepts:

* 02:00:00
* 2:00:00
* 10:00:00
* etc

Fixes Issue #239.


wwlwpd:
Turns out this flushes out another issue in asgs_main.sh that seems to document the fact that the leading 0 is required. The change to `qscript.pl` didn't cause this, but it exposed the limitation. I recommend we fix this in qscript and then figure out how to fix anywhere else it is an issue. (recommendation will remain to use a leading zero for now).

Issue #244, opened 2020-05-19
asgsh 'set' command conflicts with bash built-in


wwlwpd:
* replace 'set' cmd with 'define' because it conflicts with bash built-in
* added basic commandline support to `asgsh`
* `asgsh -h` prints commandline usage, then exits (1)
* `asgsh -p PROFILE-NAME` starts the shell and immediately loads `PROFILE-NAME`

PR #245, opened 2020-05-20
Replaced 'set' command with 'define' due to bash keyword conflict.

wwlwpd:
Issue #244: This issue was discovered when implementing commandline
flag processing to the asgsh shell command that starts the ASGS
Shell Environment. 'set' is a built-in keyword of bash and asgsh
was using it for other purposes.  The new command is 'define' and
works exactly the same as 'set' did. The documentation included
with asgsh was also updated to reflect this change.

Currently supported flags for asgsh:

-h             -> print command line synopsis, then quits to login shell
-p profileName -> set profile name to start in at the commandline  (e.g.,: % asgsh -p storm042020)

Resolves Issue #244.

### DIFF
--- a/cloud/general/DOT-asgs-brew.sh
+++ b/cloud/general/DOT-asgs-brew.sh
@@ -20,15 +20,23 @@
 
 # Developer Note:
 # This must remain consistent with what's exported by asgs-brew.pl, and
-# vice versa; if something is added in set(), be sure to add the
+# vice versa; if something is added in define(), be sure to add the
 # corresponding entry in show() - including usage
 
 help() {
+  echo Command Line Options \(used when invoking asgsh from login shell\):
+  echo "   -h                         - displays available asgsh command line flags, then exits."
+  echo "   -p     profile             - launches the ASGS Shell environment and immediate loads specified profile on start, if it exists."
   echo
-  echo Commands:
+  echo ASGS Shell Commands:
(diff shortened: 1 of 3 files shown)


wwlwpd:
Updated the quick reference guide also to v1.2, PDF is attached.
[ASGS Shell Quick Reference - v1.2 draft.pdf](https://github.com/jasonfleming/asgs/files/4654640/ASGS.Shell.Quick.Reference.-.v1.2.draft.pdf)

wwlwpd:
Most disruptive change is anyone used to using `set` now must use `define`. BUT the big win here is that starting `asgsh` can be done by specificying the profile:

`asgsh -p arthur2020`

Which is useful when opening up a separate terminal to do things like monitor the current run (e.g., `tailf syslog`).

Before, it took 2 commands and a few seconds:

* `asgsh` - start shell
* (wait a second or 2)
* `load profile arthur2020`

This save a lot of time and thinking, critical during a real time event. I plan to build on these flags to allow `asgsh` to do things currently requiring manual use of `asgs-brew.pl` - I think I can relegate that script to more of a boot strap script (used on first install then very infrequently). E.g., I can envision `asgsh` to be able to update itself - `asgsh --update` or be used to re-run any portion of the original build...anyway that's all irrelvant now but could become useful in the future.

Issue #131, opened 2019-08-31
all operators need to add their public keys to operator_defaults.sh


jasonfleming:
The config/operator_defaults.sh script contains public keys used to log into various HPC machines and copy results to THREDDS servers ... to facilitate the granting of access to these resources, all Operators should put all their public keys in config/operator_defaults.sh.

wwlwpd:
Be **very** sure not to include the private keys. I have a security concern also, but I will have to mention it private. Not that I have to give bad guys any ideas.

That asside, a default $HOME/.ssh/config file might provide some opportunity for simplifying the logic of the platform.sh or wherever the TDS servers are defined.

`Host tdd1
  Hostname <operator-specific-tdds-host-here>
  User <operator-specific-here>
  IdentityFile /path/to/<operator-specific-tdds-private-key-here>

Host tdd2
Hostname <operator-specific-tdds-host-here>
User <operator-specific-here>
IdentityFile /path/to/<operator-specific-tdds-private-key-here>

Host tdd3
Hostname <operator-specific-tdds-host-here>
User <operator-specific-here>
IdentityFile /path/to/<operator-specific-tdds-private-key-here>
`
The advantage here is all the variance is handled in the $HOME/.ssh/config, but asgs can assume that the hosts aliased by "tdd1", "tdd2", and "tdd3" above.

But for however you may utilize a standard $HOME/.ssh/config file, the option is there. I say this also because what I saw in the most recent platforms.sh reminded me very much of this nice feature of ssh and it's config file options.

hapoo:
I have added my Public Key with pull request #145

wwlwpd:
Here's my proposal:

1. we maintain an `authorized_hosts` file to which we add public keys (a single one, not differentiated by resource or site)
2. public keys should be a minimum strength, I don't know what to suggest immediately but I think it's not a bad idea to require that they be at least 4096 bits (i.e., created via `ssh-keygen -b 4096`)
3. public keys should be stripped of any identifying information, that's basically anything after the `==` at the end of the string; `ssh-keygen` typically adds in `$user@$hostname`; since github is public (it'd be much better if this was handled privately, however; but assuming there are eyes on this...)
[...]

PR #249, opened 2020-05-23
Making ASGS aware of $HOME/.ssh/config for ssh-based connections.

wwlwpd:
With the addition of an ssh config under the control of the operator,
the scripts need only to reference the Host alias as defined in
$HOME/.ssh/config. Under each stanza, any of the existing parameters
may be changed (Username, Port, etc) as well as any ssh or scp
options. Since rsync uses ssh by default, it would automatically
derive host specific settings by virtue of using the Host alias as
the reference to the remote server.

This change separates the concerns of ASGS from managing ssh usernames,
hosts, private keys, and "-o" options to focusing on using ssh connections
on a higher level for interacting with remote servers.

The change also allows operators to draw from the same basic ssh config
file, requiring them to only ensure that their respective usernames and
paths to private keys are correct. The list of servers and ssh options
can be maintained and be consistent across the entire effort and in
conjunction with resource administrators.

Resolves Issue #131.

### DIFF
--- a/output/asgsConvertR3ToNETCDF.pl
+++ b/output/asgsConvertR3ToNETCDF.pl
@@ -36,8 +36,6 @@ my $DeleteFiles=0;
 my $SendNotification=1;
 my $CopyFile=1;
 sub stderrMessage($$);
-my $sshkey;  # public key for scp'ing results to remote hosts
-my $opendapuser; # username for scp'ing results to remote hosts
 my $opendaphost; # hostname for scp'ing results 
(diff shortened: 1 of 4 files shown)


wwlwpd:
Update, I am running the latest master with this patch applied on lonestar. Once I am convinced it is working, I will update this issue.

wwlwpd:
@jasonfleming might be a good idea to get this in so I can work with the other operators to help them with their `$HOME/.ssh/config` files for the next storm.

Issue #246, opened 2020-05-20
make core scripts (e.g., asgs_main.sh) aware of some envars


wwlwpd:
If set in an `asgsh` environment, it'd be nice for the following variables to override any other settings when `asgs_main.sh` is first started:

* `ADCIRCDIR`
* `ACCOUNT`
* `SERQUEUE`
* `QUEUENAME`
* `PPN`
* `STORM`
* `YEAR`
* `COLDSTARTDATE`
* `FTPSITE`
* `RSSSITE`

The rationale would be to integrate the environment in asgsh more closely with the core scripts.

Core scripts should function as they do now if these variables were not set in the environment (no surprises).

wwlwpd:
Closing this. The main benefit was in `ADCIRCDIR`.

Going to rethink this in the model of a long running process with tunables that can be tweaked without shutting ASGS down.

Note done (not really needed at this time)
* `ACCOUNT`
* `SERQUEUE`
* `QUEUENAME`
* `PPN`
* `STORM`
* `YEAR`
* `COLDSTARTDATE`
* `FTPSITE`
* `RSSSITE`

PR #254, opened 2020-05-23
Make asgs_main.sh aware of ADCIRCDIR if set in environment.

wwlwpd:
Issue 246: Adds awareness to ASGS (asgs_main.sh) of the ADCIRCDIR
environmental variable if set, which is the case with an asgsh
profile when set up as recommended (e.g., using initadcirc and/or
load adcirc <adcirc-installation-name>).

The effective change here is that if ADCIRCDIR is set in the user's
environment when asgs_main.sh is executed, that this value is used
with out having to set it explicity in the ASGS configuratino file.
If ADCIRCDIR is not set or is set to an empty value, asgs_main.sh
will fall back to the traditional behavior of setting it to "null",
thus require ADCIRCDIR to be set explicitly in the configuration
file.

Issue #246 mentions other environmental variables ASGS should track,
but this commit only covers ADCIRCDIR since it provides immediate
value and alleviates some confusion experienced by operators who
are learning to use the shell environment.

### DIFF
--- a/asgs_main.sh
+++ b/asgs_main.sh
@@ -1483,7 +1483,7 @@ variables_init()
    HOTORCOLD=coldstart
    LASTSUBDIR=null
    FTPSITE=null
-   ADCIRCDIR=null
+   ADCIRCDIR=${ADCIRCDIR:-null} # will respect ADCIRCDIR if already sent in the environment
    SCRATCHDIR=null
    MAILINGLIST=null
    QUEUESYS=null


wwlwpd:
> This is an interesting change. I had not thought about using environment variables, but it makes sense in the context of asgsh. With so many ways to set the version of ADCIRC being used (and potentially other aspects, I wonder if we need some way of tracing provenance of variable values : platforms.sh, asgs config, environment.

There are a handful of environmental variables I'd like to manage in this way (via the environment). The are listed in #246:

- ADCIRCDIR [done]
- ACCOUNT
- SERQUEUE
- QUEUENAME
- PPN
- STORM
- YEAR
- COLDSTARTDATE
- FTPSITE
- RSSSITE

There might be more.  The change in this PR shows that `asgs_main.sh` can be made environmentally aware without sacrificing the robustness that's been built in for running outside of `asgsh`. For me, running it in the shell and saving the environments via profies is the way to go. But I don't want to break how it works outside of the shell environment either. I just don't want us to be responsible for us trying to troubleshoot "unsupported" frankenstacks. :)

> Also, where does the setting of environment variables fit in the override hierarchy?

If it's set in the environment, `asgs_main.sh` uses that value instead of `null`. So assuming `ADCIRCDIR` is not set anywhere else, the one it's initialized to in `asgs_main.sh` is it. But placing it in: the configuration script, `configs/operator_defaults.sh`, `platforms.sh`, and anywhere else along the way will over write it.

PR #257, opened 2020-05-26
All Python scripts should have proper shebang and be executable.

### LINKED ISSUE #256: ensure all Python scripts have correct shebang and are executable
Correct shebang is:

    #!/usr/bin/env python

They should all be set to executable, 0755.

Note: this is consistent with how all `perl` scripts are managed and executed.

wwlwpd:
Issue #256: Added '#!/usr/bin/env python' and made sure all .py were
chmod 0755. This will allow us to treat all scripts executable and
rely on the shebang line rather than try to manage which $PYTHON (or
$PERL, etc) we run a script with.

Resolves #256.


wwlwpd:
Additional clean up work to ensure a stable and predictable python environment.

PR #259, opened 2020-05-26
Removed reliance on RMQMessaging_Python envar, purged platforms.sh

### LINKED ISSUE #258: remove reliance on or need to set RMQMessaging_Python
This is a rather aggressive change and builds on #256:

* removes need to set RMQMessaging_Python (retained locally for some scripts, but set with `$(which python)`.
* removes loading of any python related modules in `platforms.sh`

Note: this is consistent with how all `perl` scripts are managed and executed.

wwlwpd:
Issue #258: This is a rather aggressive change, but one that is needed
for python stability. It removes the need for "RMQMessaging_Python" to
be set and is not used to run any scripts since all .py scripts should
have the proper shebang and be executable as a result of #256. The loading
of Python via the environment tool "module" that is common on HPCs has
also been removed to further minimize any confusion about which `python`
should be used to execute any scripts.

Resolves #258.


wwlwpd:
Note: This depends on #256/#257 to be resolved/merged.

PR #261, opened 2020-05-28
Added regession test for monitoring/FortCheck.py

### LINKED ISSUE #260: add regression test for FortCheck.py (and smallish fort.61.nc)
test is to be run during an `asgsh` `verify` command and after the building of `python`

wwlwpd:
Issue #260: Added a smallish fort.61.nc for FortCheck.py to read;
this is a regression test to ensure that the Python environment in
the ASGS Shell Environment (asgsh) is working as expected.

If the test fails, it indicates a problem with how the environment
finds the right python executable to run or indicates issues with
the installation of one of the Python modules.

Resolves #260.


wwlwpd:
I added this to provide some confidence that the asgsh Python environment was set up correctly.

PR #309, opened 2020-06-25
Fixed init-asgs.sh guess for supermic.

### LINKED ISSUE #308: fix supermic guess for default platform in init-asgs.sh
returning `smic` instead of `supermic`, which is expected by `platforms.sh`

wwlwpd:
Issue #308: Was guessing as "smic". Should be "supermic" based on
what's defined in platforms.sh.

Resolves #308.


wwlwpd:
merging, tested

PR #356, opened 2020-07-30
Issue 342 - move user specific stuff to ~/.asgsh_profile

wwlwpd:
I attempted to move all the user specific bits into user specific asgs_profile files each user to move over or at least use as a good starting point. I also turned config/operator_defaults.sh into a wrapper around source'ing $HOME/.asgsh_profile if it exists since removing it would have meant updating a ton o folder ASGS configs.

Let me know if there are any obvious problems. I think I want to let this one bake a little like we did with the ssh_config change.

It's 2 commits. Commit 1 is editing platforms.sh. Commit 2 is dealing with operator_defaults.sh.


wwlwpd:
this PR is still valid, just bumping activity on it

jasonfleming (approved):
Hey Brett these look great, will be nice to have this rationalized going forward in 2021. If you haven't already, can you tag the repo before merging this pull request as 2020stable? The merging of this PR will be the beginning of the 2021 cycle.

wwlwpd:
Thanks!

PR #384, opened 2020-09-24
Moving some asgsh bash functions to shell scripts in ./bin

### LINKED ISSUE #105: install all to support FigureGen
build and install gdal, gmt, shorelines data file, and imagemagick (if necessary); all to support FigureGen

wwlwpd:
Issue #105: Moved all bash functions that don't affect the environment
of the parent process (asgsh) into external scripts located in ./bin.

Resolves #105.


jasonfleming:
I like this modularization!

PR #394, opened 2021-01-12
Use different backtrace option based on compiler.

wwlwpd:
Issue 393: Intel compilers were not able to find the "backtrace"
routine, rather it needs the "tracebackqq" routine - gfortran does
use the "backtrace" routine. A define flag was added and is set
when the it is detected that "ifort" is being used to compile all
of Fortran code (or most of it) in ASGS.

Resolves Issue 393.


wwlwpd:
closing this in favor of one that includes the image-magick support

PR #396, opened 2021-01-22
Gnuplot etc

wwlwpd:
Adds:

* `gnuplot` (nox)
* `ffmpeg`
* `units`
* updates Dockerfiles, make sure `zip` is installed on Ubunut (via Dockerfile)


jasonfleming (approved):
Brett this looks great. Makes a lot more sense than adding it to the Dockerfile, like I was thinking. Please go ahead and merge.

Issue #391, opened 2020-12-23
enable automated building of adcirc v54 and adcirc v55 in asgs brew


jasonfleming:
This is a prerequisite for running with the latest version(s) of adcirc in the 2021 season. This enhancement should wait until we have tagged the 2020stable version of asgs.

wwlwpd:
This is great. We're going to have to figure out what to do about `cmake` and establish a more formal workflow with upstream ADCIRC development. As part of this I'd prefer to tie in supporting the makefile infrastructure, at least for ASGS needs. Quick look across platforms shows that cmake is available via `modules` but that's a risky proposition as we've seen before. Frankly, I'd rather commit to maintaining the makefile infrastructure. Anyway, discussion is needed on that part.

wwlwpd:
Current tact is to go with v55 only, testing on queenbee in v55 the
make command to "make unhcat" fails:

  make: *** No rule to make target `unhcat'.  Stop.

Will update comments with more information as we proceed on QB.

wwlwpd:
Another error,

```
ifort -c -O2 -W0 -assume byterecl -traceback -diag-disable 8290 -diag-disable 8291 -diag-disable 8293  -I /work/asgs/adcirc-cg-v55release-intel/work/odir33/ -
o /work/asgs/adcirc-cg-v55release-intel/work/odir33/swanpre2.o  /work/asgs/adcirc-cg-v55release-intel/thirdparty/swan/swanpre2.f
Cannot disable Fortran error message 7002
/work/asgs/adcirc-cg-v55release-intel/thirdparty/swan/swanpre2.f(1031): error #7002: Error in opening the compiled module file.  Check INCLUDE paths.   [SWN_O
UTNC]
      USE swn_outnc                                                       41.52
----------^
compilation aborted for /work/asgs/adcirc-cg-v55release-intel/thirdparty/swan/swanpre2.f (code 1)
make[1]: *** [/work/asgs/adcirc-cg-v55release-intel/work/odir33/swanpre2.o] Error 1
make[1]: Leaving directory `/work/asgs/adcirc-cg-v55release-intel/work'
make: *** [adcswan] Error 2
```

jasonfleming:
Thank you Brett, that is interesting, I will give it a shot too so I can reproduce it. Let's keep a record of the platforms where we see this and those that we don't (if any).

wwlwpd:
Skipping v54, going straight for v55. Seems the only breakages
[...]

PR #409, opened 2021-03-09
Added support build and "profile" for ADCIRC v55release.

wwlwpd:
Issue #391 - Adding support for building v55 of ADCIRC and
updating support for the "load adcirc" command (which updates
$PATH, etc based on where the binaries are expected to be upon
successful compilation)

NOTE: This commit doesn't represent an exhaustive validation process
for use of v55 in ASGS operationally. But this commit is a
precondition for such a validation.


jasonfleming (approved):
This looks good Brett, thank you.

PR #423, opened 2021-03-30
Added workaround for ImageMagick build issue, first noticed on Debian.

wwlwpd:


jasonfleming (approved):
This looks good to me. I like the addition of the Debian Dockerfile with a dual purpose of acting as a reference for installing directly onto host OS. The ImageMagick changes are outside my area of expertise, so I am an interested bystander on that. :-)

Issue #447, opened 2021-04-09
make qscript.pl add -R to command line options to padcirc and padcswan when asgs HOTSTARTCOMP=subdomain


jasonfleming:

wwlwpd:
thanks for assigning this to yourself, lets me know you intend to work on it; don't hesitate to assign me to stuff - this is actually helpful for me know know expectations :)

wwlwpd:
I recommend we get this merged before #450 is started, or conversely include #450 as part of this work, just as a separate commit.

Issue #444, opened 2021-04-08
add flex to docker containers to support building of nco


jasonfleming:

wwlwpd:
Is this required for you on ubuntu for NCO?

If it is, I recommend we add it as a package installation in the Ubuntu
and Debian (and future Alpine) Dockerfiles. (e.g., `apt-get install -y flex`).

If you're seeing this on an HPC platform, the I agree it probably should be
supported in `asgs-brew.pl`.

wwlwpd:
Disregard. Done.

PR #466, opened 2021-04-27
Issue 442

### LINKED ISSUE #442: add support for ec2001 and openwater meshes
These meshes are to be used for benchmarking and testing the adcirpolate workflow.

jasonfleming:
Resolves #442 #444 and #447


wwlwpd (commented):
This looks fine, it's a lot of commits so I created
a tag in `master` called, `pre-merge-issue-442` in case
we need to go back for some reason. I don't think we'll
need it.

PR #474, opened 2021-05-04
Removed all references to JOBENV and JOBENVDIR

### LINKED ISSUE #293: evaluate need for "JOBENV" section of qscript.template
@jasonfleming - I noticed this section in `qscript.template` that could potentally destabilize the batch environment if we're going to go with a clean room approach.

```
# source scripts to set required PATH and LD_LIBRARY_PATH
declare -a JOBENV
JOBENV=%jobenv%
for script in ${JOBENV[*]}; do
   source %jobenvdir%/$script
done
```

We may need to determine that this can be safely removed - or it could provide a more precise way to update the batch environment than the use of the `module` command that's currently under some (my) scrutiny.

jasonfleming:
Resolves #293


wwlwpd (commented):
Changes look fine, but can you confirm that you meant to change 28 files? Thanks for making this all 1 comment :)

PR #494, opened 2021-05-18
includes metadata.pl for supporting run.properties, json, and yaml files 

jasonfleming:
includes some rearrangements of utilities in the directory structure


metadata.pl
```diff
@@ -0,0 +1,273 @@
...
+# a set of key-value pairs.
+#
+# Alternatively, it will accept a string of key-value pairs to
+# be added to the metadata file.   
+#
+# The script determines the format of the metadata file 
+# automatically using the file extension (.json, .yaml, or 
+# .properties).
+#--------------------------------------------------------------
+$^W++;
```
wwlwpd, line 33: Just for future reference, this can be accomplished with `use warnings;` - just to make it look similar to `use strict;`.

metadata.pl
```diff
@@ -0,0 +1,273 @@
...
+   notification.email.job_failed_list notification.hpc.email.notifyuser
+   notification.opendap.email.opendapnotify 
+);
+# the following properties from run.properties are stored as individual
+# top level properties but should be considered a sublist of related 
+# metadata with a list of properties associated with them in turn
+# (to start off with, " File Name" and " Format")
+# adcirc.file.output.list
+my @outputfile_properties = ( 
+    "Water Surface Elevation Stations", "Water Surface Elevation",
```
wwlwpd, line 84: (another FYI) - Since there is no variable interpolation, you can use single quotes - or alternatively wrap the strings inside of `q{ }`.

metadata.pl
```diff
@@ -0,0 +1,273 @@
...
+    "Maximum Mean Wave Period", "Peak Wave Period",
+    "Maximum Peak Wave Period", "Initially Dry", "Inundation Time",
+    "Maximum Inundation Depth", "Ever Dried", "End Rising Inundation",
+    "Wind Velocity 10m Stations", "Wind Velocity 10m",
+    "Maximum Wind Speed 10m",
+);
+
+our $this = "metadata.pl"; 
+#
+GetOptions(
```
wwlwpd, line 101: (FYI) - I've been using `GetOptions` like this more:

```
my %opts = ();
my @flags = (qw/
  metadatafile=s
  keys=s
  mapscalar=s
  jsonify
/);
my $ok = GetOptions(\%opts, @flags);
```
Take a look at `cloud/general/asgs-brew.pl` around line 120 for a real example.

metadata.pl
```diff
@@ -0,0 +1,273 @@
...
+           "metadatafile=s" => \$metadatafile,    
+           "keys=s" => \$keys,
+           "mapsacalar=s" => \$mapscalar,
```
wwlwpd, line 104: @jasonfleming - is the flag here supposed to be `mapsacalar` [sic] or `mapscalar`?

metadata.pl
```diff
@@ -0,0 +1,273 @@
...
+# determine path to the metadata file (if any was provided)
+my $dirpath = "";
+$last_slash = rindex($metadatafile,"/");
+if ( $last_slash != -1 ) {
+    $dirpath = substr($metadatafile,0,$last_slash);
+}
+#
+#  J S O N
+if ( $type eq ".json" ) {
+    unless (open(F,"<$metadatafile")) {
```
wwlwpd, line 127: suggestion:

```
open my $F, q{<}, $metadatafile || die qq{Failed to open '$metadatafile': $!.\n};
```

metadata.pl
```diff
@@ -0,0 +1,273 @@
...
+        &stderrMessage("ERROR","Failed to open '$metadatafile': $!.");
+        die;
+    }
+    # slurp the file contents into a scalar variable
+    $file_content = do { local $/; <F> };
+    close(F);
```
wwlwpd, line 133: The slurp is actually pretty close to a good idiom I just recently learned. Update to use `$F`, and use `scalar` to ensure proper variable context:

```
$file_content = do { local $/; scalar <$F> };
close $F;
```

metadata.pl
```diff
@@ -0,0 +1,273 @@
...
+    }
+} else {
+    &stderrMessage("ERROR","File type was not recognized (requires either 'json' or 'yaml' or '.properties').");
+}
+
+1;
+
+#
+# write a log message to stderr
+sub stderrMessage () {
```
wwlwpd, line 243: @jasonfleming - next time might want to look into `Log::Log4perl` - it's not currently installed by default, but would not be a bad addition to our set of modules. It makes logging with multiple output targets a little more manageable.

wwlwpd (changes requested):
@jasonfleming - there's a potential typo in your argument handling code, also had a couple of suggestions in there as well. Otherwise, looks good to me. The additional modules install no problem using `cpanm` so they're in the right place, e..g., `./PERL-MODULES`.

wwlwpd:
Updated and merged.

PR #500, opened 2021-05-19
    New run.properties messaging for run properties database.

BrianOBlanton:
Added ASGS pid to runprop message
    Added new property called forcing.metclass with values of tropical|synoptic.


wwlwpd (approved):
looks good

PR #506, opened 2021-06-01
Expanded ASGS env variables save by asgsh shell profiles.

### LINKED ISSUE #504: expand environment saved by `asgsh` profiles
Somewhat related to #503, it's handy to have the "WALLTIME" variables saved when they need to necessarily deviate from what's set in `config/mesh_defaults.sh` (currently).

wwlwpd:
Issue 504: Now saving the following as part of profile:

    *HINDCASTWALLTIME
    *ADCPREPWALLTIME
    *NOWCASTWALLTIME
    *FORECASTWALLTIME
    *QUEUENAME
    *SERQUEUE
    *ACCOUNT
    *PPN
    *INTENDEDAUDIENCE
    *USERIVERFILEONLY
    *RIVERSITE
    *RIVERDIR
    *RIVERUSER
    *RIVERDATAPROTOCOL
    *FTPSITE

In addition to just saving these in the asgsh profile, asgs_main.sh
is also aware that they may be set in the environment. As a result,
when the following variables are set in the environment (managed by
the profile), those values are used in lieu of the existing defaults.

Resolves #504.


jasonfleming (approved):
Hey @wwlwpd this looks great, thank you this looks great. One question that I have is: what is your preferred approach to the Operators changing these values while the ASGS is running? (a) stop asgs, e.g., export NOWCASTWALLTIME=10:00:00, and then start asgs? (the values of these variables are set from environment in variables_init() which is executed only upon startup), or (b) change the value in the asgs config file for that asgs instance, and wait for the asgs to re-read its config file?  Or something else?

wwlwpd:
> Hey @wwlwpd , thank you this looks great. One question that I have is: what is your preferred approach to the Operators changing these values while the ASGS is running? (a) stop asgs, e.g., export NOWCASTWALLTIME=10:00:00, and then start asgs? (the values of these variables are set from environment in variables_init() which is executed only upon startup), or (b) change the value in the asgs config file for that asgs instance, and wait for the asgs to re-read its config file? Or something else?

Likely option _a_; and this is also why profiles can now save them. I suspect the defaults will be fine for now and it'll probably rare to export them in asgsh itself. Since it's set in asgs_main.sh by default, it won't affect the environment unless the operator explicitly exported this.

PR #517, opened 2021-06-07
fixed empty cpurequest property

### LINKED ISSUE #516: fix undefined cpurequest property so asgs uses right QUEUENAME on QBC
When running on QBC the cpurequest property is empty (at least in a 1st nowcast when hotstarting from URL) which erroneously causes the queue script to try to run large parallel jobs in the "single" queue instead of "workq".

jasonfleming:
Also fixed QUEUENAME setting to workq for parallel jobs requesting more than 48 cores on QBC. Resolves #516

### DIFF
--- a/asgs_main.sh
+++ b/asgs_main.sh
@@ -1220,10 +1220,8 @@ submitJob()
    #
    CLOPTIONS=""     # command line options
    LOCALHOTSTART=""
-   CPUREQUEST=$NCPU
    if [[ $NUMWRITERS != "0" ]]; then
       CLOPTIONS="-W $NUMWRITERS"
-      CPUREQUEST=`expr $NCPU + $NUMWRITERS`
    fi
    # record the number of requested CPUs for use in determining capacity to run another job
    if [[ $HOTSTARTCOMP = subdomain ]]; then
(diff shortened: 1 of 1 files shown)


asgs_main.sh
```diff
@@ -1293,6 +1291,7 @@ submitJob()
       DATETIME=`date +'%Y-%h-%d-T%H:%M:%S'%z`
       echo "time.${JOBTYPE}.start : $DATETIME" >> run.properties
       echo "[${DATETIME}] Starting ${JOBTYPE}.${ENSTORM} job in $PWD." >> ${ADVISDIR}/${ENSTORM}/${JOBTYPE}.${ENSTORM}.run.start
+      CPUREQUEST=`expr $NCPU + $NUMWRITERS`
```
wwlwpd, line 1294: This looks good.

In `bash` I like to use the following idiom for adding or subtracting since it saves on the creation of a new process,

    CPUREQUEST=$(($NCPU + $NUMWRITERS))

wwlwpd (approved):

asgs_main.sh
```diff
@@ -1754,6 +1753,7 @@ writeJobResourceRequestProperties()
    logMessage "$THIS: Writing properties associated with compute job to $1/run.properties."
    # on queenbeeC, if a parallel job uses 48 or fewer cores, it
    # should be submitted to the single queue to avoid "low utilization" emails
+   CPUREQUEST=`expr $NCPU + $NUMWRITERS`
    if [[ $HPCENV = "qbc.loni.org" && $CPUREQUEST -le 48 ]]; then
```
wwlwpd, line 1757: I think next time we're in this part of the code, turning this into a `case` statement would probably a good move (based on `$HPCENV`.

PR #539, opened 2021-06-22
Audited placement of the use/capture of $? (exit code).

### LINKED ISSUE #538: make test for job errors in queue script occur immediately after command being tested
One thing for sure is capturing `$?` must occur immediately after the command, e.g.:

image (6)

Here you've got 2 assignments before, which is going to reset `$?` to 0, so that right there is indicating a potentially false success, which means it might be an actual error, but this explains why `padcirc.finish` is getting written.

wwlwpd:
Issue 538: Changes ensure clarity or that $? is used/captured immediately
after the command of concern.

Resolves #538.


jasonfleming (approved):
This looks great, thank you.

PR #556, opened 2021-07-09
Removed echohelp function; it is no longer needed.

### LINKED ISSUE #555: deprecate echoHelp() function from asgs_main.sh
This function can be removed.

jasonfleming:
resolves #555


wwlwpd (approved):
Looks great, thank you. I am a big fan of identifying and deleting old code.

Issue #507, opened 2021-06-01
RMQ* environment needs to be rolled up into a bash associative array


wwlwpd:
Currently in `asgs_main.sh`, the following are declared (below a number of `declare -a` arrays):

```
# RMQMessaging defaults
   RMQMessaging_Enable="off"   # "on"|"off"
   RMQMessaging_Transmit="off" #  enables message transmission ("on" | "off")
   RMQMessaging_Script="${SCRIPTDIR}/monitoring/asgs-msgr.py"
   RMQMessaging_Script_RP="${SCRIPTDIR}/monitoring/rp2json.py"
   RMQMessaging_StartupScript="${SCRIPTDIR}/monitoring/asgs-msgr_startup.py"
   RMQMessaging_NcoHome="/set/RMQMessaging_NcoHome/in/asgs/config"
   namedot=${HPCENVSHORT}.
   RMQMessaging_LocationName=${HPCENV#$namedot}
   RMQMessaging_ClusterName=$HPCENVSHORT
```
These should be rolled up into a bash array, examples of these are used throughout the code using the `declare -a` keyword/option, e.g.:

```
   declare -A RMQMessaging
   RMQMessaging[Enable]="off"   # "on"|"off"
   RMQMessaging[Transmit]="off" #  enables message transmission ("on" | "off")
   RMQMessaging[Script]="${SCRIPTDIR}/monitoring/asgs-msgr.py"
   RMQMessaging[Script_RP]="${SCRIPTDIR}/monitoring/rp2json.py"
   RMQMessaging[StartupScript]="${SCRIPTDIR}/monitoring/asgs-msgr_startup.py"
   RMQMessaging[NcoHome]="/set/RMQMessaging_NcoHome/in/asgs/config"
   namedot=${HPCENVSHORT}.
   RMQMessaging[LocationName]=${HPCENV#$namedot}
   RMQMessaging[ClusterName]=$HPCENVSHORT
```

BrianOBlanton:
@wwlwpd so do you mean to put these into an "init_rmq.sh" and put init_rmq.sh into something like 
declare -a MESSAGING_INIT=()
?

wwlwpd:
Actually, I am sorry I will update the issue. What I was really asking for was an associative array. 

https://riptutorial.com/bash/example/6886/associative-arrays

So, something like

```
   declare -A RMQMessaging
   RMQMessaging[Enable]="off"   # "on"|"off"
[...]

PR #572, opened 2021-07-16
Cleaning up anything relying on RMQMessaging_Enable - first pass.

wwlwpd:


wwlwpd:
There are a couple of key things that I addressed in this PR

1. One had to do with a cat command that was getting $CONFIG content to send to RMQ in the start up, If $CONFIG is hosed some how (e.g., file doesn't exist) on start up, the cat command will hang - low probability but possible

2. The signal traps in monitoring/logging.sh were calling RMQ functions before the RMQ functions were defined in the same file; this is bash so best to only call things after they've been clearly defined

Those both could have caused "weird" issues. I also added some checks for RMQ scripts that were assumed to exist and quoted anything going into the actual Python based client.

wwlwpd:
Any concerns with this PR @BrianOBlanton ?

wwlwpd:
Once this gets through, I'll follow up with a PR for #507.

jasonfleming (approved):
These are all great. We have our work cut out for us, in terms of getting things cleaner and tighter throughout ASGS. This is a nice step in that direction.

PR #577, opened 2021-07-19
Localized "$THIS" where possible, inside of bash function.

### LINKED ISSUE #573: add local keyword to THIS assignment in bash functions
There is a lot of clutter in our bash functions when they are trying to add their names to their log messages, due to all of them using THIS. This parameter should be declared local within functions to avoid having to change the value back after the end of the function.

wwlwpd:
Issue 573: Added 'local' keyword to assignment to "$THIS" in all
bash functions. When used globally, set "THIS=(basename -- $0)",
which provides consistent naming of "$THIS" across all scripts.

Furthermore, there's no need to reset or "change back" $THIS when
a function is called. This took care of itself when a script that
defined $THIS globally in the script file exited because a child
process can't affect the environment of the parent.

Resolves #573.

### DIFF
--- a/archive/enstorm_pedir_removal.sh
+++ b/archive/enstorm_pedir_removal.sh
@@ -23,7 +23,8 @@
 # This script gets all the information it needs from the
 # run.properties file instead of using command line arguments.
 #----------------------------------------------------------------
-THIS=archive/enstorm_pedir_removal.sh
+THIS=$(basename -- $0)
+
 # this assumes this script is executed in the dirctory that is to be archived
 SCENARIODIR=$PWD
 SCRIPTDIR=`sed -n 's/[ ^]*$//;s/path.scriptdir\s*:\s*//p' run.properties`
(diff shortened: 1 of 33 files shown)


wwlwpd:
@jasonfleming - marked as a draft because this really needs to be scrutinized. I was able to start ASGS with no change to functionality or output, which was encouraging. But did not test all changes files since this would require quite a few full runs and cycles. Let me know what you think.

jasonfleming (approved):
Yes sorry about the awkward `WASTHIS`es ... I was hazy on scoping in bash. Did some reading and now have a better understanding. I did not know about `THIS=$(basename -- $0)`, that is a great idea. Based on this I just did some googling and found the bash builtin array `FUNCNAME` which always contains the call stack. We could use that instead of manually setting this ourselves for use in error messages. Let's keep this in draft, I'd like to take a shot at making the changes to use `FUNCNAME` if that works for you.

wwlwpd:
Yes sound good to me. Let me know if this falls off for you and you'd like me to take care of that also. It might also not be a bad idea to get this in and worked out, then look at `$FUNCNAME` in a follow up PR.

wwlwpd:
Please create a new issue for `$FUNCNAME` - good idea, better to do after we get past this and the JSON status change.

jasonfleming:
Totally agree. I will create a new issue for the FUNCNAME thing. Please go ahead and merge this one.

PR #581, opened 2021-07-22
Modify OpenDAP notification email if 'nowcast' or 'hindcast'.

### LINKED ISSUE #568: unique email subject line for NOWCASTS
To avoid any potential conflicts with ASGS run postings to CERA, would it be possible to please create a unique email subject line for NOWCASTS? While CERA is currently not on the email notification list, that would prevent any accidental postings to CERA if a notification should go out in the future.

E.g., can the subject line be changed from

ADCIRC POSTED for 2021071400-1 qbc.loni.org. SABv20a_nam_jgf 
to 

ADCIRC NOWCAST POSTED for 2021071400-1 qbc.loni.org. SABv20a_nam_jgf

wwlwpd:
Issue 568: If posting to OpenDAP and $SCENARIO is a 'nowcast' or
'hindcast', add it to the subject line.

Resolves #568.

### DIFF
--- a/output/opendap_post.sh
+++ b/output/opendap_post.sh
@@ -212,10 +212,29 @@ for server in ${SERVERS[*]}; do
    opendapEmailSent=no
    #
    runStartTime=${properties["RunStartTime"]}
+
+   # dev NOTE: we should not get crazy here with all of the options of
+   # passing metadata via subject; the following does little but gives
+   # an indication of the chaos that can ensue if we perpetuate this
+   # method - if email subject modification is requested again, be sure
+   # bring it up for discussion during team deliberations
+
+   # default primary subject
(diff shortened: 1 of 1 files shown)


jasonfleming (approved):
I agree that this type of change introduces too much coupling between sender and receiver. For example, I am anticipating having multiple nowcast scenarios, including at least a nowcastWind10m and a nowcast with data assimilation. Both would require modifying this conditional. Basically we would start developing a list of acceptable scenario names for the subject line. I think the solution (in the offseason) is to get rid of email as a messaging solution for results notification. I see this change as a (necessary) emergency workaround.

carolakaiser:
Looks good to me. Thank you, Brett!

wwlwpd:
Thanks, @carolakaiser and @jasonfleming!

PR #583, opened 2021-07-22
changes for real time status monitoring of asgs via json on opendap

jasonfleming:
This is extremely draft at this point, but wanted to get this out there in case there is interest in having an advanced look.


wwlwpd:
I'll look more later, but this is a great change. It represents a major code change. I'd like to get all current PRs merged (after review, any fixups) before this PR is "official" so we properly rebase _this_ change from `master` after (not the other way around).

jasonfleming:
Thank you, that is great to hear. Totally agree that this is a big step and that the other PRs are way more important at the moment.

PR #588, opened 2021-07-24
removing "-assume buffered_io"

wwlwpd:
Note, @jasonfleming, these patches are applied to the following ADCIRC patch sets:

* v53release-issue-549
* v55release-issue-549

I figured we could batch them into this, since we're using these to test the supermic fixes. Once
we finally get it licked, I can roll the accumulated patches into the patch sets for the other respective
versions of adcirc


jasonfleming (approved):
This looks great @wwlwpd I hope this helps with stability in the coupled model. Looking back at Matt's initial email, I wonder what platform he was running on, since we've had so much weird stuff with SuperMIC. Looking at the actual functioning of the `-assume buffered_io` in the Intel documentation, I think this will also give us some immediacy in our logging, at the possible expense of i/o slowdowns, which may be worth it if it helps in troubleshooting.

PR #589, opened 2021-07-24
Addressing issue with stdout redirected to 'null' in logMessage.

wwlwpd:


jasonfleming (approved):
Thank you for finding and fixing this.

PR #591, opened 2021-07-24
Put HPC QUEUENAME into a "hints" function in ./platforms.sh

wwlwpd:


wwlwpd:
Moved this into _draft_ to do some testing while setting up the tx2008 on frontera

wwlwpd:
This seems to be working as expected on frontera, so putting it back into "ready for review" status.

jasonfleming (approved):
This is a great approach @wwlwpd thank you. There is one more wrinkle here that may or may not matter: priority queue (or partition, or qos). Submitting small jobs to a priority queue may or may not improve throughput. But in any case, each platform has idiosyncracies on this that may be best addressed by adding at least one more argument to the list supplied to this shell script that represents the "vipxxx" or "priority" aspect.

wwlwpd:
> This is a great approach @wwlwpd thank you. There is one more wrinkle here that may or may not matter: priority queue (or partition, or qos). Submitting small jobs to a priority queue may or may not improve throughput. But in any case, each platform has idiosyncracies on this that may be best addressed by adding at least one more argument to the list supplied to this shell script that represents the "vipxxx" or "priority" aspect.

Good suggestion! Will take a look at that; basically the idea is the hide all the ikcy logic in this method. So considering `QOS` is certainly in the realm of _acceptable_. The trick is just making sure it's passed using quotes, so if it's not defined then the positional arguments (`$1`, etc) won't be affected - it'll just be empty.

PR #603, opened 2021-07-29
Add parameter for `"$QOS"`used at TACC to `HPCQueueHints`

wwlwpd:


jasonfleming (approved):
This looks great @wwlwpd. It looks to me like the interpretation of QOS will be different on different platforms inside HPCQueueHints.

PR #604, opened 2021-07-29
Abstracting hint for PPN for serial jobs on queenbee, supermic.

wwlwpd:
Work in progress.


jasonfleming (approved):
Love it. I did not like the little hacks I made here and there to accommodate these things, but did not know what else to do. Its great to have these neatly encapsulated in `platforms.sh`.

jasonfleming (commented):
This is great, thank you @wwlwpd ... does this (or should this) include our mechanism for setting the vip qos at TACC?

wwlwpd:
> This is great, thank you @wwlwpd ... does this (or should this) include our mechanism for setting the vip qos at TACC?

hmmm not sure what you mean, but let's talk about it - if we can set other things this way, I am all for it; in the mean time I'll merge this

PR #612, opened 2021-08-02
Moved RMQ variable initialization 

### LINKED ISSUE #611: initialization of rmq messaging in asgs_msgr.sh fails because script name is empty
The default values of the RMQMessaging variables are set in `variables_init()` but only if the RMQMessaging is enabled, but the `variables_init()` function is executed before the `platforms.sh` file or asgs config file is read.

jasonfleming:
RMQ init moved from `asgs_main.sh->variables_init()` to `monitoring/asgs-msgr.sh` to resolve an issue where these variables were not getting set because their initial default values were set only if messaging was enabled, and this was performed before the config file was read to see if the Operator had enabled the messaging. Resolves #611


wwlwpd (approved):
Looks good. Note, this change means that only `RMQMessaging_Enable` and `RMQMessaging_Transmit` are available "globally" stemming from the parent `asgs_main.sh` _parent_ process. The others will not persist after the child prcess of `monitoring/asgs-msgr.sh`. finishes.

PR #620, opened 2021-08-14
Removed advisory directories with .old extension from list of directories to be used in validation

### LINKED ISSUE #619: validation workflow picks up old advisory directories that should not be used
The `nco_ncrcat.pl` script which is used to facilitate the validation workflow picks up in-situ directories with a `.old` extension that indicates that they are not to be used.

jasonfleming:
Resolves #619

### DIFF
--- a/output/nco_ncrcat.pl
+++ b/output/nco_ncrcat.pl
@@ -82,7 +82,7 @@ GetOptions(
 # hindcast scenarios for use in validation
 my @files = `find $root_data_dir -name $datafile -print | grep -E 'nowcast|hindcast' | sort`;
 foreach my $f (@files) {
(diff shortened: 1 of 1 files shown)


output/nco_ncrcat.pl
```diff
@@ -82,7 +82,7 @@
 # hindcast scenarios for use in validation
 my @files = `find $root_data_dir -name $datafile -print | grep -E 'nowcast|hindcast' | sort`;
 foreach my $f (@files) {
-   #stderrMessage("DEBUG","$f");
+   &stderrMessage("DEBUG","$f");
```
wwlwpd, line 85: `&` is superfluous, but it's fine.

wwlwpd (approved):
looks good, merging

PR #623, opened 2021-08-19
Issue 622

BrianOBlanton:
PR for transmit_rps fix

### DIFF
--- a/monitoring/logging.sh
+++ b/monitoring/logging.sh
@@ -217,6 +217,9 @@ RMQMessageRunProp()
 { 
   if [[ ${RMQMessaging_Enable} == "off" || ! -e "${RMQMessaging_Script_RP}" ]]
   then
+    estr="warn: RMQMessageRunProp did not send rp message because either RMQMessaging_Enable is off or RMQMessaging_Script_RP not found. "
+	echo $estr
+	allMessage $estr
     return 1
   fi
 
@@ -236,6 +239,13 @@ RMQMessageRunProp()
      --Transmit "${RMQMessaging_Transmit}"         \
      --input_filename "$RPDIR/run.properties"      \
      --output_filename "$RPDIR/run.properties.json" >> ${SYSLOG} 2>&1
+
+  if [ $? == 0 ] ; then
+    allMessage "${RMQMessaging_Script_RP} returned a 0 status."
+  else
+    allMessage "${RMQMessaging_Script_RP} returned a non-0 status, with these parameters: $ASGS_PID, $INSTANCENAME, $RPDIR, ${RMQMessaging_Transmit}" 
+  fi 
+
 }
 
 #  send message when shutting down on INT and clear all processes
--- a/output/transmit_rps.sh
+++ b/output/transmit_rps.sh
@@ -27,10 +27,11 @@ SCENARIODIR=$PWD
 RUNPROPERTIES=$SCENARIODIR/run.properties
 
 if [ -e "$RUNPROPERTIES" ] ; then
-        echo "Found run.properties file"
+    echo "Found run.properties file"
+    allMessage "transmit_rps.sh found the run.properties file."
 else
-        echo "run.properties not found file"
-        exit 1
+    allMessage "transmit_rps.sh did NOT find the run.properties file."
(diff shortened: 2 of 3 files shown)


monitoring/logging.sh
```diff
@@ -236,6 +239,13 @@ RMQMessageRunProp()
      --Transmit "${RMQMessaging_Transmit}"         \
      --input_filename "$RPDIR/run.properties"      \
      --output_filename "$RPDIR/run.properties.json" >> ${SYSLOG} 2>&1
+
```
wwlwpd, line 242: If you capture `$?` with a variable, then you can use it later. E.g.,
```
  esta=$?
  if [ $esta == 0 ] ; then
    allMessage "${RMQMessaging_Script_RP} returned a 0 status."
  else
    allMessage "${RMQMessaging_Script_RP} returned an error status ($esta), with these parameters: $ASGS_PID, $INSTANCENAME, $RPDIR, ${RMQMessaging_Transmit}" 
  fi 
```

output/transmit_rps.sh
```diff
@@ -26,12 +26,15 @@ declare -A properties
...
+# get ASGS functions
+SCRIPTDIR=`sed -n 's/[ ^]*$//;s/path.scriptdir\s*:\s*//p' $RUNPROPERTIES`
+source $SCRIPTDIR/properties.sh
+source $SCRIPTDIR/monitoring/logging.sh
+source $SCRIPTDIR/platforms.sh
+
+if [ ! -e "$RUNPROPERTIES" ] ; then
     allMessage "transmit_rps.sh did NOT find the run.properties file."
-    return 1
+    exit 1
```
wwlwpd, line 37: Looking into this, I am a little confused how this script is being called since I am seeing conflicting implication in the commits. (e.g., `exit` is concerning because if it's not called in a sub-shell, it will end the `asgs_main.sh` process).

wwlwpd: @jasonfleming can you provide some insight?

BrianOBlanton: this is called in the post processing hook, which I thought was a sub-shell process.

monitoring/logging.sh
BrianOBlanton: OK

output/transmit_rps.sh
jasonfleming: I checked out the latest `transmit_rps-fix` to see how `transmit_rps.sh` was being executed ... this script is one element of the $POSTPROCESS bash array of post processing scripts, and they each get executed one at a time as separate commands. Based on my understanding of the bash process model, this script should execute as its own process, and the `exit` command should only kill the `transmit_rps.sh` process. 
The post processing as a whole (as well as other processes that occur after a forecast scenario has been submitted) is executed within a subshell, i.e., 
```
( 
   ... 
) &
```
so if the exit did kill more than just the `transmit_rps.sh` (which I'm thinking it won't) it could at most kill the forecast scenario subshell. Not sure how this was executed in previous commits, but this is what I see in the latest on this branch.

jasonfleming (commented):
This looks ok to me. The real fix here is in the parsing of the run.properties file in the `rp2json.py` script. The other changes in this PR are troubleshooting support, which is also important. I think this workflow could be simplified via refactoring, but such refactoring is outside the scope of this pull request.

output/transmit_rps.sh
wwlwpd: FWIW, that `&` causes an asynchronous child process; should we get rid of it (will still be a sub shell) or put a `wait` after it?

wwlwpd: I know that's out of scope for this, but just something to note.

jasonfleming: Hey @wwlwpd this is meant to be asynchronous, since the forecast scenarios are "fire and forget" ... the ASGS submits them and then leaves them with the job monitoring, post processing, and archiving code they are supposed to run, and just leaves them to it goes on to the next task (or next advisory cycle).

wwlwpd: I am going to merge this so we can have it for Henri

PR #633, opened 2021-08-29
Added HPC_Reservation_Hint() function 

### LINKED ISSUE #632: when a reservation is available on Frontera, jobs submitted to the small queue should not use it
Reservations can be granted on Frontera, but these reservations distinguish between the `normal` queue and the `small` queue, and a reservation created for the `normal` queue will cause a job submitted for that reservation in the `small` queue to not get scheduled.

jasonfleming:
to take care of picking whether not to submit job to a reservation depending on platform and number of CPUs requested. Resolves #632

### DIFF
--- a/asgs_main.sh
+++ b/asgs_main.sh
@@ -729,10 +729,14 @@ prepFile()
    unset _PPN
 
    if [[ $QUEUESYS = "SLURM" ]]; then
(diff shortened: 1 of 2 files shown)


wwlwpd (approved):
Looks good!

PR #662, opened 2021-09-28
Replace unsupported strftime with external `date` call in awk script.

### LINKED ISSUE #661: awk on some platforms do not have the strftime function
I am seeing error messages like the following on some platforms
```
awk: /work/asgs/monitoring/scenarioStatus.awk: line 63: function strftime never defined
```
Not sure what the best solution would be. It might make more sense to replace the offending awk scripts with perl scripts (as opposed to making sure we have access to gawk across all platforms). Please advise.

wwlwpd:
Issue 661: Uses external call to `date` with "getline" capture of output
in awk to capture formatted date. Improves portability of script.

Resolves #661.

### DIFF
--- a/monitoring/scenarioStatus.awk
+++ b/monitoring/scenarioStatus.awk
@@ -54,7 +54,8 @@ END {
       # need to write the final job data
       print "{ \"jobtype\" : "jobtype" \"time.submit\" : "submit" \"jobid\" : "jobid" \"time.start\" : "start" \"time.finish\" : "finish" \"time.error\" : "error" }"
       print "]," # end of jobs.status array
-      date=strftime("%Y-%m-%d-T%H:%M:%S%z")
+      # replace strftime with direct call to `date`
+      "date +%Y-%m-%d-T%H:%M:%S%z" | getline date
       print "\"time.scenario.status.lastupdated\" : \""date"\""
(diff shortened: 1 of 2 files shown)


jasonfleming:
I have to confess that I do not understand why this fix works, but I can confirm that it does. :-) I have also duplicated this fix in the `timestamp.awk` script, because it was suffering from the same issue. @wwlwpd can you have a look?

jasonfleming (approved):
It works! Thank you :-)

wwlwpd:
> I have to confess that I do not understand why this fix works, but I can confirm that it does. :-) I have also duplicated this fix in the `timestamp.awk` script, because it was suffering from the same issue. @wwlwpd can you have a look?

```
 "date +%Y-%m-%d-T%H:%M:%S%z" | getline date
```

Is to `awk`, what the following is to the shell:

```
date=$(date +%Y-%m-%d-T%H:%M:%S%z)
```

It's a little confusing because the first `date` in the fix is the call to `date` in a subshell; the second `date` is the variable that gets used in the following line. In `awk`, `getline` is a little more explicit for getting the output of `date +%Y-%m-%d-T%H:%M:%S%z`.

wwlwpd:
Thanks for applying it to the other script.

PR #670, opened 2021-10-04
Removed NOTIFYUSER from main scripts.

### LINKED ISSUE #666: Question:  what's the difference among `OPENDAPNOFITY=..` and `NOTIFYUSER=...`  and `statusNotify=...`
@jasonfleming - I am wondering if we can collapse these down to 1; at the very least I'd like to avoid any other _notification_ type lists. Any thoughts?

wwlwpd:
Issue 666: Removed from on bash function, which required the
shifting of positional parameter processing; then removed it
from the call to this function.

Resolves #666.


jasonfleming (commented):
Hey @wwlwpd this is great, I didn't realize it would be so straightforward to get rid of `NOTIFYUSER`. However, when looking through the code, I see we have a few more "notify" related parameters: `ASGSADMIN`, `NOTIFYUSER`, and `NOTIFY_SCRIPT`. This seems like a good time to revisit these too. I will add this as a question to #666 so we can exchange ideas on this.

jasonfleming:
Oops in the above comment, I meant to say `EMAILNOTIFY` rather than `NOTIFYUSER`.

wwlwpd:
@jasonfleming - I captured this in the new issue, #677.

PR #674, opened 2021-10-07
Changes for supermic, letsencrypt root CA fallout.

wwlwpd:
Issue 667: Fixing a number of issues noticed when trying to build
on supermic (LSU) recently. In addition to their being some issues
with downloading certain Perl modules from CPAN over port 80 http,
there are internet wide issues with the way letsencrypt chose to
expire their root CA. SSL is useless in general, so just turning that
off for downloading all sources used to build using asgs-brew.pl that
require https URLs and wget to fetch the files.

Resolves 667.

### DIFF
--- a/PERL-MODULES.wget
+++ b/PERL-MODULES.wget
@@ -1,0 +1,1 @@
+Archive::Zip https://www.cpan.org/authors/id/P/PH/PHRED Archive-Zip-1.68 tar.gz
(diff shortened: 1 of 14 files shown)


wwlwpd:
build ok on:

* qbc
* qb2
* supermic

testing tacc but don't expect any surprises

PR #681, opened 2021-10-12
[testing in progress] Remove any mainline use of system mail or sendmail.

### LINKED ISSUE #673: RFC: eliminate reliance on system `sendmail` altogether, support only `asgs-sendmail.pl`
(RFC = _request for comments_)

@jasonfleming - what do you think, this would offer some benefits for notification reliability _and_ reduce some code complexity (noticed in `opendap_post.sh`).

wwlwpd:
Issue 673: Removed OPENDAPMAILSERVER since it's not needed, and
removed any use of "mail" utility on the current system in favor
of asgs-sendmail.pl.

Resolves #673.

This is a proof of concept to show what it'd look like.

### DIFF
--- a/PERL-MODULES
+++ b/PERL-MODULES
@@ -23,3 +23,4 @@ Archive::Zip
 JSON::XS
 YAML::Tiny
 DBD::SQLite
+Util::H2O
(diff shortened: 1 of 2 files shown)


wwlwpd:
> but it will continue in the case of `ASGSADMIN`

Do you mean it gets added to the queue script and we rely on the batch system to send the email? Or do you mean we call `mail` somewhere inside of the queue script directly outside of the batch system?

PR #690, opened 2021-10-21
Allow OPENDAPNOTIFY to be updated in config during runtime.

### LINKED ISSUE #685: (verify) make sure that we can modify the `OPENDAPNOTIFY` list in the ASGS config file and have the change be gracefully picked up by ASGS without a restart.
One thing is that we want to make sure that we can modify the `OPENDAPNOTIFY` list in the ASGS config file and have the change be gracefully picked up by ASGS without a restart. I'm not sure what implications this has (if any) for the suggested workaround.

_Originally posted by @jasonfleming in https://github.com/StormSurgeLive/asgs/issues/609#issuecomment-942668025_

wwlwpd:
Issue 685: OPENDAPNOTIFY was getting derived from run.properties
once set in the initial run via the opendap_post.sh script(s). This
change removes this propagation via run.properties and allows it
to be changed at run time during the configuration file, which is
preferred.

Resolves #685.

### DIFF
--- a/output/opendap_post.sh
+++ b/output/opendap_post.sh
@@ -71,8 +71,6 @@ if [[ ${#FILES[@]} -eq 0 ]]; then
    warn "cycle $CYCLE: $SCENARIO: $THIS: No files to post to opendap servers in $RUNPROPERTIES."
    exit
 fi
-OPENDAPNOTIFY="${properties['notification.opendap.email.opendapnotify']}"
-#
 if [[ $SCENARIO == "asgs.instance.status" ]]; then
     statusDir=${properties['path.statusdir']}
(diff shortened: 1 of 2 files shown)


wwlwpd:
Correct, it's like anything you update in the configuration file - with this change it'll at least get picked up the next go around - but yeah, other than editing `run.properties` directly before `opendap_post.sh` is run. But you got it right, the nature of the scripts make it a delayed update of the new value.

wwlwpd:
@jasonfleming - any more thoughts on this one?

wwlwpd:
@jasonfleming - bump

jasonfleming:
Hey @wwlwpd I am open to this, can we discuss on our call tomorrow? The OPENDAPNOTIFY stuff is an absolute mess that needs to be rebuilt in any case. The only issue here is that I don't fully understand this particular PR. I just need to talk through it a bit.

jasonfleming (approved):
Yes! :-)

PR #695, opened 2021-11-02
Added 'jq' as an option post-ASGS-install build step via asgsh.

### LINKED ISSUE #692: add `jq` to build/install step of `asgs-brew.pl`
https://stedolan.github.io/jq/

>jq is like sed for JSON data - you can use it to slice and filter and map and transform structured data with the same ease that sed, awk, grep and friends let you play with text.
> jq is written in portable C, and it has zero runtime dependencies. You can download a single binary, scp it to a far away machine of the same type, and expect it to work.

wwlwpd:
Issue 692: Extended asgsh's 'build' command to support 'jq'; this
commit is also a demonstration of a path forward towards offering
optional things that don't necessarily need to be added at ASGS
build time. In the future, most new software requests should be
available via what will become a poor man's "package manager".

Comands supported by 'build':

  build adcirc             # invokes ADCIRC patchset wizard
  build jq [clean|rebuild] # standard approach moving ahead
  build avail              # (lists available things to build)

Resolves #692.

### DIFF
--- a/cloud/general/DOT-asgs-brew.sh
+++ b/cloud/general/DOT-asgs-brew.sh
@@ -753,20 +753,23 @@ echo " --update-shell option"
 echo
 fi
 
-# runs script to install ADCIRC interactively
-build () {
-  if [ -z "${1}" ]; then
-    echo "The 'build' command requires an argument specifying what to build, e.g., 'build adcirc'"
-    exit
-  fi
+# 'build' is basically ASGS Shell Environment's "package manager"
+# these are the "optional" installs - from individual utilities
+# to "bundles" (e.g., a set of related, but optional Perl modules)
+build() {
   TO_BUILD=${1}
-  case "${1}" in
+  BUILD_OPTS=${2}
+  case "${TO_BUILD}" in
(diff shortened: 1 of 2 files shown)


jasonfleming (approved):
This looks good, and I am looking forward to learning more about `jq`. I guess the reason why it is a "poor man's" package manager is that we aren't looking at dependencies between features in an automated way, which is a good thing. I like the idea of a flexible installer with core packages and optional add-ons; I think this facilitates container-based packaging as well.

wwlwpd:
> This looks good, and I am looking forward to learning more about `jq`. I guess the reason why it is a "poor man's" package manager is that we aren't looking at dependencies between features in an automated way, which is a good thing. I like the idea of a flexible installer with core packages and optional add-ons; I think this facilitates container-based packaging as well.

Great, thank you. If we _really_ need an embedded package manager, there's something called `pkgsrc` that's used on a lot of OSes. It's the base package system for NetBSD, but has grown into its own thing - https://www.pkgsrc.org/ (not suggesting it, just pointing it out).

Issue #724, opened 2022-01-26
bad numpy version required by netCDF4 python 2.7 module is breaking installation


wwlwpd:
Immediate issue - ASGS build is broken.

Current best option:

* deprecated python2 (or provide it as an optional install after ASGS is already built/installed)
* add build and environmental support for python3 (minimally to support RMQ/ADCIRC dashboard)

wwlwpd:
@BrianOBlanton - looking to deprecated python2 and support python3. How much time would you need to make sure your RMQ scripts work with python3 once I make it available?

wwlwpd:
Fixed in #725. Other issues split off into #726, #727.

PR #725, opened 2022-01-27
Downgraded python netCDF4

wwlwpd:
Issue 724: Downgraded netCDF4 python module to 1.1.7.1 because
this is the newest version of the module that requires a supported
dependency on numpy.

Part of #724.


wwlwpd:
Merging since this issuing blocking installation on all platforms.

PR #728, opened 2022-01-27
Adding optional build support for minimal python 3.7.

### LINKED ISSUE #727: add build and environmental support for python3
Adding only as an optional install (via `asgsh`'s `build python3`) - no plans to automatically install it, but that could change

wwlwpd:
Issue 727: Python 3.7.10 may now be installed into the ASGS Shell
Environment with the command, `build python3`. It will not conflict
with python (2.7.18). Python 2 will be deprecated before the coming
hurricane season. Also, moving forward python will not be installed
automatically unless it's deemed necessary.

Note: No non-core modules are installed, but `pip3` is installed. If
there are python3 modules that should be included by default, they
will be added to this script when identified.

Resolves #727.

### DIFF
--- a/cloud/general/DOT-asgs-brew.sh
+++ b/cloud/general/DOT-asgs-brew.sh
@@ -766,10 +766,18 @@ build() {
     jq)
       init-jq.sh ${ASGS_INSTALL_PATH} ${BUILD_OPTS}
       ;;
+    python2)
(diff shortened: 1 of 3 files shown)


jasonfleming (approved):
Hey @wwlwpd thank you for this. I am not really qualified to review it, since I am a python dabbler, but it looks good to me!

PR #721, opened 2022-01-17
extended default output end time to 9999.0 days

jasonfleming:
Resolves #720

Issue #719, opened 2022-01-12
Image::Magick perl module test fails when brewing perl 5.34 perl environment 


jasonfleming:
When brewing the perl environment for perl 5.34 on Stampede2 and Frontera, I observed this error from the `verify` command: 
```
1..27
not ok 1 - require Image::Magick;
#   Failed test 'require Image::Magick;'
#   at /work2/00976/jgflemin/stampede2/asgs/cloud/general/t/verify-image-magick.t line 7.
#     Tried to require 'Image::Magick'.
#     Error:  Can't locate Image/Magick.pm in @INC (you may need to install the Image::Magick module) (@INC contains: /work/00976/jgflemin/stampede2/opt/default/perl5/perls/perl-5.34.0/lib/site_perl/5.34.0//x86_64-linux /work/00976/jgflemin/stampede2/opt/default/perl5/perls/perl-5.34.0/lib/site_perl/5.34.0/ /work2/00976/jgflemin/stampede2/asgs/PERL /work/00976/jgflemin/stampede2/opt/default/perl5/perls/perl-5.34.0/lib/site_perl/5.34.0/x86_64-linux /work/00976/jgflemin/stampede2/opt/default/perl5/perls/perl-5.34.0/lib/site_perl/5.34.0 /work/00976/jgflemin/stampede2/opt/default/perl5/perls/perl-5.34.0/lib/5.34.0/x86_64-linux /work/00976/jgflemin/stampede2/opt/default/perl5/perls/perl-5.34.0/lib/5.34.0) at (eval 8) line 2.
1..1
# Looks like you failed 1 test of 1.
```
This is not a major issue because I don't know that any of our perl scripts are using this module (yet). Just wanted to flag it for documentation.

wwlwpd:
Will try to reproduce

wwlwpd:
@jasonfleming I could not reproduce with perl-5.34.0 on Frontera, but I did take the opportunity to update to the currently latest IM - 7.1.0-22. I will try to reproduce on stampede2. Maybe next time you're on Frontera try to rebuild ASGS?

jasonfleming:
Yes, @wwlwpd I think that is the next step for me. Thank you for working on this.

PR #729, opened 2022-02-03
Updated Image-Magick from 7.1.0-9 to 7.1.0-22.

wwlwpd:
Issue 721: We were not able to reproduce the issue described in
the problem report (#721), but took the opportunity to updatre
to the latest version of Image-Magick, currently 7.1.0-22.

Relates to #721.


jasonfleming:
I think there was a hiccup on the related issue, #719 ... other than that, this looks great, it is actually good that it could not be reporduced ... I will do a clean install on Frontera.

jasonfleming (approved):
Thank you for attempting to reproduce this. Good to see the version update.

Issue #723, opened 2022-01-20
ASGS not generating padcirc.slurm file?


BrianOBlanton:
Trying to get the latest ASGS/ASGSH running.  On initialize/hindcast, get this log message: 

`asgs_main.sh>submitJob()>hindcast : Submitting /projects/ncfs-dev/ncv99-nam-bob-2022/asgs23353/initialize/hindcast/padcirc.slurm.`

but jobErr contains:  `sbatch: error: Unable to open file padcirc.slurm`

and padcirc.slurm does not exist.   The usual log files (padcirc.out, adcirc.log, fort.16) are all empty.

BrianOBlanton:
Can you add me to some list? 

ERROR: Permission to StormSurgeLive/asgs-configs.git denied to BrianOBlanton.

BrianOBlanton:
[ncfs-dev-ncv99_nam.sh.txt](https://github.com/StormSurgeLive/asgs/files/7935663/ncfs-dev-ncv99_nam.sh.txt)

wwlwpd:
weird, will have to look into why this is required

wwlwpd:
@BrianOBlanton added you to `asgs-configs` - let me know if you can't get things uploaded there, no need to do a branch or anthing; just commit to `master`

wwlwpd:
I get around to trying to run this on ht4, ty

wwlwpd:
@BrianOBlanton - this might be related to #733 - can you most the header of the `partmesh.slurm` file that was generated for one of the failed runs? I can confirm is the latest commit to `master` should fix it.

BrianOBlanton:
this is the slurm parameter section at the top of prep15.slurm. 

#SBATCH --job-name="prep15.hindcast"
#SBATCH --time=120
#SBATCH --ntasks-per-node=1
#SBATCH --ntasks=1
#SBATCH --nodes=1
#SBATCH --partition=batch
#SBATCH --qos=null
#SBATCH --output=/projects/ncfs-dev/ncv99-nam-bob-2022/asgs23353/initialize/hindcast/prep15.out
#SBATCH --mail-type=FAIL,TIME_LIMIT

BrianOBlanton:
here is the partmesh.slurm file:

[partmesh.slurm.txt](https://github.com/StormSurgeLive/asgs/files/8004656/partmesh.slurm.txt)

jasonfleming:
[...]

PR #734, opened 2022-02-04
Queue script was getting illegal value for failed email.

### LINKED ISSUE #733: deprecated run.property was getting injected into queue scripts with illegal values
Related to @akheir's issues on qbc.

wwlwpd:
Issue 733: updated qscript.pl to inject the propery email address/
property into the queue script. Issue found while debugging issue
on QBC when running partmesh phase. May be related to bblanton's
issue on ht4, #723 (maybe, need to confirm).

Resolves #733.


wwlwpd:
kind of a hot fix and tested to work, may need to follow up with a better property but this was a reasonable guess I think...

PR #735, opened 2022-02-05
Added python 3.7.9 + modules + verification, completely removed python2

### LINKED ISSUE #730: add python3 to `asgs-brew.pl`, remove python2 from `asgs-brew.pl` - keep for now as optional build via `build python2`
replaces #726

wwlwpd:
Issue 730: Added Python 3.7.9 as a replacement for Python 2.7.18 (EOL),
3.7 installed as part of standard ASGS installation. Pythone 2 was removed
just because it was going to be deprecated anyway, but due to it getting
unmaintained online (as well as the compatible modules), it was not
going to be work fixing just to deprecate.

The latest "pika" module for RabbitMQ is installed, but not tested.

The latest netCDF4 module is also installed, and the old netCDF4 test
script works without modification. So that's both the asgs-brew verification
step and also what's run in asgsh when "verify" or "verify python" is
commanded.

No python scripts were removed, but if anyone relies on them or wishes
to use them, they will likely have to be updated to work with Python
3.7.

Resolves #730.


jasonfleming (approved):
This looks great @wwlwpd. I hope it wasn't too much of a headache. It's nice to have this implemented so we are prepared for the future!

PR #744, opened 2022-02-10
Issue 743 - ADCIRC patches to support rostam

wwlwpd:


wwlwpd:
tested builds on rostam and supermic, all worked fine

PR #763, opened 2022-03-08
Based date math for NAM data on UTC timezone

### LINKED ISSUE #762: specify UTC as timezone in bash date commands
When using the `date` command in `bash`, the utility defaults to the local time zone, which may include daylight savings time. In cases where this should be avoided, the UTC time zone should be explicitly specified.

jasonfleming:
resolves #762 

I am testing this now on QBC.


jasonfleming:
This is running on QBC.

wwlwpd (approved):
Looks good to me! The only concern I'd have is portability, but we use almost exclusively GNU `date` (as opposed to BSD `date` which has some minor differences where POSIX standard seems to allow).

PR #764, opened 2022-03-08
Issue 738 - proper isolation among 2 or more ASGS installations on the same machine

wwlwpd:


jasonfleming (dismissed):
Hey @wwlwpd this may be considered a separate issue, but I am looking at line 205 of the modified `init-asgs.sh` 
```
if [[ -z "$quit" || "$quit" = Y ]]; then
```
and wondering if the string comparison could be made case insensitive? Sometimes when we are under pressure we might just mash `y` and not realize. Looking at this stackoverflow

https://stackoverflow.com/questions/1728683/case-insensitive-comparison-of-strings-in-shell-script

it looks like we can get this by changing these lines by e.g.
```
if [[ -z "$quit" || "${quit^^}" = Y ]]; then
```
Overall these changes look great, even a simplification over previous behavior in some cases. Thank you.

jasonfleming:
@wwlwpd One question I have about this is how we should migrate our existing instances to this new isolated structure. Or should we just blow away and reinstall?

wwlwpd:
1. Nice catch on the string comp, will add
2. for "updating" old installations, I would suggest doing the following before building out the 2nd+ installations:

* `cd /work/$USER/asgs` # existing installation
* `cp $HOME/bin/asgsh .`
* `cp $HOME/bin/update-asgs .`

Once that happens and you start building out the new isolated instances, there won't be any confusion about what to do to start `asgsh` or update things.

wwlwpd:
Read for review, did the following:

* used the kind of check you found for all y/n (regardless of the default answer)
* "default" profile suggestion is much more friendly for remembering what `SCRIPTDIR` this is for

During the review, if you have not, I suggest using this branch to try to build an additional installation, even if you already have doe this to get the full effect.

jasonfleming:
Can do, thank you. I will install this on my desktop machine to test.

jasonfleming:
Ok I have tested this on my desktop machine successfully. Now I will try cloning a fresh copy of ASGS and testing out a different branch side-by-side (also on my desktop).

wwlwpd:
Gonna use privilege and merge this to get it out the queue. I've tested it and it seems to work as advertised.

PR #771, opened 2022-03-18
Catch exception thrown when Net::FTP timesout during `ls`.

### LINKED ISSUE #770: clean up "Timeout" messages in `bin/get_nam_status.pl`
`Net::FTP` was throwing some uncaught exceptions and making it hard to debug issues while porting to Rostam

wwlwpd:
Issue 770: Wrapped $ftp->ls() in an eval {} and now outputs a meaningful
message; also exits script with an error code of 1. All $ftp calls in
this script were wrapped with eval {}, all "die"s were changed to "exit
1" in order to keep the logs cleaner. Explicitly set Timeout => 120
(seconds), which is the default - just to make it more clear.

Resolves #770.

### DIFF
--- a/bin/get_nam_status.pl
+++ b/bin/get_nam_status.pl
@@ -68,39 +68,40 @@ if ( $startcycle eq "null" && $jshash_ref ) {
    }
 }
 #
-ASGSUtil::appMessage(
-          "DEBUG",
-          "Connecting to $backsite:$backdir");
+ASGSUtil::appMessage( "DEBUG", "Connecting to $backsite:$backdir");
 our $dl = 0;   # true if latest status was determined successfully
 # open ftp connection
-our $ftp = Net::FTP->new($backsite, Debug => 0, Passive => 1);
-unless ( defined $ftp ) {
-   ASGSUtil::stderrMessage("ERROR",
-                           "ftp: Cannot connect to $backsite: $@");
-   die;
+our $ftp = Net::FTP->new($backsite, Debug => 0, Passive => 1, Timeout => 120);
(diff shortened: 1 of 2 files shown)


wwlwpd:
`ls` timeout errors now look roughly like this,

```
2022-Mar-18-T01:32:52-0500 ERROR: get_nam_status.pl: ftp: Cannot list "all" files: [Net::FTP] Timeout                                                         
2022-Mar-18-T01:35:54-0500 ERROR: get_nam_status.pl: ftp: Cannot list "all" files: [Net::FTP] Timeout                                                         
2022-Mar-18-T01:38:55-0500 ERROR: get_nam_status.pl: ftp: Cannot list "all" files: [Net::FTP] Timeout                                                         
2022-Mar-18-T01:41:56-0500 ERROR: get_nam_status.pl: ftp: Cannot list "all" files: [Net::FTP] Timeout  
```

wwlwpd:
2nd commit is for #772, will not squash on merge

jasonfleming (approved):
Thank you for these reliability and troubleshooting improvements, they look great and I learned something new! I wonder if the JSON slurp/decode combo should be abstracted into a sub in the ASGSUtil pm. I also wonder why this is timing out on Rostam.

wwlwpd:
> wonder if the JSON slurp/decode combo should be abstracted into a sub in the ASGSUtil pm

I don't thing so, Perl is one of those languages that benefits from "idioms" that are often easier to use and read than shoving in a function somewhere. This is an example of where that's just how you tend to do it after having done it a bunch :)

PR #779, opened 2022-03-19
Added a ASGS config "linter", as part of a preflight check.

### LINKED ISSUE #756: `asgs-lint` command/script - startup health check and configuration echo at start of ASGS
When the ASGS starts, it performs a limited health check. It does not check allocation status for example. The combination of the existing ASGS startup tests and the `verify` checks from `asgsh` could be used right at the start of an ASGS startup to catch issues prior to execution. This could even be done as a dry run, to check configuration, existence of files, availability of allocation, etc.

wwlwpd:
Issue 756:

Resolves #756.

### DIFF
--- a/bin/asgs-lint
+++ b/bin/asgs-lint
@@ -1,0 +1,724 @@
+#!/usr/bin/env perl
+
+use strict;
+use warnings;
+
+package bin::asgslint;
+
+use Util::H2O::More qw/h2o opt2h2o/;
+use Getopt::Long qw//;
+
+our $VERSION = q{1.0.0};
+
+use constant {
+    EXIT_SUCCESS => 0,
+    EXIT_ERROR   => 255,
+};
+
+#--
+#    TRIGGER must be ftp, rss, rssembedded, or atcf.
+#    NCPUCAPACITY must be greater than or at least equal to NCPU plus NUMWRITERS for each individual scenario.
+#    BACKGROUNDMET set to something other than allowable values : on, NAM, OWI, namBlend, and off.
+#    BACKGROUNDMET parameter can only be set to off or namBlend when TROPICALCYCLONE is set to something other than off.
+#    TROPICALCYCLONE and BACKGROUNDMET cannot both be on simultaneously
+#    TROPICALCYCLONE and BACKGROUNDMET cannot both be off simultaneousl
+#    LASTSUBDIR cannot be set to null in the config file if the ASGS is trying to hotstart (i.e., HOTORCOLD=hotstart), and the STATEFILE does not exist.
+#    LASTSUBDIR cannot be set to null in the STATEFILE file if it exists and the ASGS is trying to hotstart (i.e., HOTORCOLD=hotstart).
+#Done, Optional, if defined
+#    VORTEXMODEL parameter set to something other than SYMMETRIC, ASYMMETRIC, and GAHM.
+#    QUEUESYS must be serial, mpiexec, SLURM, or PBS.
+#    VORTEXMODEL parameter set to something other than GAHM when BACKGROUNDMET set to namBlend.
+#    if VARFLUX=on or VARFLUX=default for any ADCIRC version,
+#    If WAVES=on and SWANDT is sent in config, it must be in the range, 300 to 1200
+#    If WAVES=on and REINITIALIZESWAN is set in config, it must be either 'on' or 'off'
+#    If WAVES=on, adcswan and padcswan executables must be available via ADCIRCDIR (exported to environment via ADCIRC config)
+#    If WAVES=on, unhcat.exe executables must be available via SWANDIR (exported to environment via ADCIRC config)
+#    If WAVES=on, swaninit.template and $SWANTEMPLATE files exist in input/meshes/common/swan under either $SCRIPTDIR
+#    ADCIRC ENV checks
+#TODO
+#    If VARFLUX=default then verify that the file $INPUTDIR/$RIVERFLUX exists and has nonzero length
+#    if VARFLUX=on or VARFLUX=default and the ADCIRC version is less than v53, then verify that the file $INPUTDIR/$RIVERINIT exists and has nonzero length
+#    If VARFLUX=on and USERIVERFILEONLY=no then verify that
+#    RIVERDATAPROTOCOL is set to one of scp, ftp, or filesystem
+#    if RIVERDATAPROTOCOL=scp then verify that the ssh server $RIVERSITE is reachable as the user $RIVERUSER and that the directory $RIVERDIR exists and is readable
+#    if VARFLUX=on or VARFLUX=default for any ADCIRC version,
+#    verify that the file $INPUTDIR/$HINDCASTRIVERFLUX exists and has nonzero length
+#--
+
+my $CONFIG_VARS = {
+    INSTANCENAME => {
+        ord   => 1,
+        req   => 1,
+        check => [],
+        hint  => q{Must be defined, no spaces},
+    },
+    ASGSADMIN => { ord => 5, req => 0, check => [], hint => q{must be an email address}, },
+    PPN       => {
+        ord   => 10,
+        req   => 0,
+        check => [ \&_int ],
+        hint  => q{Must be integer greater than 0},
+    },
+    NCPU => {
+        ord   => 15,
+        req   => 1,
+        check => [ \&_int ],
+        hint  => q{Must be integer greater than 0},
+    },
+    NUMWRITERS => {
+        ord   => 20,
+        req   => 1,
+        check => [ \&_one ],
+        hint  => q{Must be integer greater than 0 (1 is recommended)},
+    },
+    NCPUCAPACITY => {
+        ord   => 25,
+        req   => 1,
+        check => [ \&_int ],
+        hint  => q{Must be integer greater than 0},
+    },
+    QUEUENAME => {
+        ord   => 30,
+        req   => 0,
+        check => [],
+        hint  => q{Name of the queue used for parallel compute jobs},
+    },
+    SERQUEUE => {
+        ord   => 35,
+        req   => 0,
+        check => [],
+        hint  => q{Name of the queue used for single processor jobs},
+    },
+    GRIDNAME      => {
+        ord => 40,
+        req => 1,
+        check => [],
+        hint => q{name of grid to use},
+    },
+    BACKGROUNDMET => {
+        ord   => 45,
+        req   => 1,
+        check => [ sub { _in( @_, qw/on off NAM OWI namBlend/ ) } ],
+        hint  => q{Must be: 'on', 'off', 'NAM', 'OWI', 'namBlend'},
+    },
+    TROPICALCYCLONE => {
+        ord   => 50,
+        req   => 1,
+        check => [],
+        hint  => q{Turn tropical cyclone forcing 'on' or 'off'},
+    },
+    TRIGGER => {
+        ord   => 55,
+        req   => 0,
+        check => [ sub { _in( @_, qw/ftp rss rssembedded atcf/ ) } ],
+        hint  => q{Must be: 'ftp', 'rss', 'rssembedded', or 'atcf'},
+    },
+    WAVES => {
+        ord   => 60,
+        req   => 1,
+        check => [ sub { _in( @_, qw/on off/ ) } ],
+        hint  => q{Turns SWAN coupling 'on' or 'off'},
+    },
+    VARFLUX => {
+        ord   => 65,
+        req   => 0,
+        check => [ sub { _in( @_, qw/on off default/ ) } ],
+        hint  => q{Turns on river flux boundary conditions support in ASGS and ADCIRC.},
+    },
+    INTENDEDAUDIENCE => {
+        ord   => 70,
+        req   => 1,
+        check => [],
+        hint  => q{Control type of notifications, 'dev', 'general', ...},
+    },
+    POSTPROCESS => {
+        ord   => 75,
+        req   => 1,
+        check => [],
+        hint  => q{List of post processing scripts},
+    },
+    OPENDAPNOTIFY => {
+        ord   => 80,
+        req   => 0,
+        check => [],
+        hint  => q{list of *emails notified when OPENDAPPOST is run via POSTPROCESS},
+    },
+    OPENDAPPOST => {
+        ord   => 80,
+        req   => 0,
+        check => [],
+        hint  => q{POSTPROCESS script used by THREDDS integration (used by monitoring/logging.sh)},
+    },
+    NOTIFY_SCRIPT => {
+        ord   => 85,
+        req   => 0,
+        check => [],
+        hint  => q{Script to handle notifications},
+    },
+    TDS => {
+        ord   => 90,
+        req   => 0,
+        check => [],
+        hint  => q{List of thredds servers to upload output},
+    },
+    COLDSTARTDATE => {
+        ord   => 95,
+        req   => 1,
+        check => [],
+        hint  => q{Start date for ramping model from zero state},
+    },
+    HINDCASTLENGTH => {
+        ord   => 100,
+        req   => 1,
+        check => [],
+        hint  => q{Length of time used to cold start},
+    },
+    HOTORCOLD => {
+        ord   => 105,
+        req   => 1,
+        check => [ sub { _in( @_, qw/coldstart hotstart/ ) } ],
+        hint  => q{Must be: 'coldstart' or 'hotstart'},
+    },
+    LASTSUBDIR => {
+        ord   => 110,
+        req   => 1,
+        check => [],
+        hint  => q{Used when HOTORCOLD is 'hotstart'.},
+    },
+    SCENARIOPACKAGESIZE => {
+        ord   => 115,
+        req   => 1,
+        check => [],
+        hint  => q{Determines what scenarios from the package to run.},
+    },
+    QUEUESYS => {
+        ord   => 120,
+        req   => 0,
+        check => [ sub { _in( @_, qw/mpiexec SLURM PBS/ ) } ],
+        hint  => q{If defined in config, must be: 'mpiexec', 'SLURM', or 'PBS'.},
+    },
+    VORTEXMODEL => {
+        ord   => 125,
+        req   => 0,
+        check => [ sub { _in( @_, qw/SYMMETRIC ASYMMETRIC GAHM/ ) } ],
+        hint  => q{If defined in config, must be: 'SYMMETRIC', 'ASYMMETRIC', or 'GAHM'.},
+    },
+    SWANDT => {
+        ord   => 130,
+        req   => 0,
+        check => [ \&_defer_check ],                                                             # defered until fatal_configcheck_WAVES_if_on
+        hint  => q{If defined and WAVES=on, reasonable values are integers between 300-1200.},
+    },
+    REINITIALIZESWAN => {
+        ord   => 135,
+        req   => 0,
+        check => [ \&_defer_check ],                                                             # defered until fatal_configcheck_WAVES_if_on
+        hint  => q{If defined and WAVES=on, must be: 'no' or 'yes'.},
+    },
+    SWANTEMPLATE => {
+        ord   => 140,
+        req   => 0,
+        check => [ \&_defer_check ],                                                                                             # defered until fatal_configcheck_WAVES_if_on
+        hint  => q{If WAVES=on, swaninit.template and $SWANTEMPLATE files must exist in $SCRIPTDIR/input/meshes/common/swan.},
+    },
+    EMAILNOTIFY => {
+        ord   => 145,
+        req   => 1,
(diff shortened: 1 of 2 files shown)


wwlwpd:
RE: swan, I can check `WAVES` and only check for `padcswan` in `PATH` if it's _on_. But the way it works is that the ADCIRC _profile_ created during an ADCIRC build is used to update `PATH`. So anything in `work/` and built for SWAN are available via `PATH` as the exist _in place_. Hope that makes sense.

wwlwpd:
> Do we want to support manually built adcirc executables that are in adcircdir but not PATH?

It doesn't matter how `PATH` is updated, really; the ADCIRC support is the easiest way to do this - but if it's not covering a situation for you, let me know. Manually build ADCIRCs can be supported by manually adding an ADCIRC "profile" in `$HOME/.adcirc-meta`.

jasonfleming:
Can the `asgs-lint` script just `source` the asgs config script to load the parameter values? Also, would you mind if I also contribute to this branch?

wwlwpd:
A linter is by definition _static_ - plus I looked into this, and it'd call a bunch of other stuff. A `source` is really just executing the script by the parent shell process, which is how it can update things in the ENVIRONMENT. So as tempting as that is, I would not recommend it as a solution.

As far as _contributing_, you are always welcome! 😄

jasonfleming:
Thank you, I was wondering about this. Given the "layered" nature of the way we do ASGS configuration at run time, dynamically, and with the optional inclusion of environment variables as well as `~/.asgs_profile` etc, I wonder if maybe the config checker needs to be dynamic to get the full picture? Not sure how to do that without changing the environment though.

jasonfleming:
Perhaps we can run the config checker in a subshell so it won't contaminate the environment?

bin/asgs-lint
```diff
@@ -62,17 +62,39 @@ if [ $(egrep -c "^ *HINDCASTLENGTH *= *[1-9][0-9]*.[0-9]+ *$" $ASGS_CONFIG) != 1
...
 ## 20
-### check BACKGROUNDMET and TROPICALCYCLONE, fatal if both are 'on'; warn if one is NOT 'on' 
+### check BACKGROUNDMET and TROPICALCYCLONE, fatal if both are 'on'; warn if one is NOT 'on'
 if [[ "$BACKGROUNDMET" == "on" && "$TROPICALCYCLONE" == "on" ]]; then
   fatal "BACKGROUNDMET and TROPICALCYCLONE can't both be 'on'. One must be 'on', the other must be 'off'."
 fi
 
+### check formatting of COLDSTARTDATE and make sure it has valid values
+### ref. https://github.com/StormSurgeLive/asgs/issues/360
+fatalMsg=$(
```
wwlwpd, line 72: I need to fix the way we extract values statically. Is `COLDSTARTDATE` being extracted?

Regarding the check, this check doesn't seem to be scaling well. I mean I don't want to force a challenge to do these kinds of things in the most difficult way possible, so I need to think about this.

My current thinking is to "warn" if it's not using `$(get-coldstart-date)` rather than validate the actual date value - this seems like this is more appropriately handled either in the running code - or - we start creating helper check scripts.

jasonfleming: Yeah this is not an optimal way to check COLDSTARTDATE, I am just playing with ideas really. The question is: in what context will the linter be executed? As ASGS is running? Inside `asgsh` but not when the ASGS is running? Outside `asgsh`? Not sure we've established this.

jasonfleming: Also, is it checking just the config file? Or also the environment variables specified in the profile?

wwlwpd: I always have them in my config files. I am thinking about redoing this in Perl as just as an experiment to see if it scales better - I think it should.

wwlwpd: And to be clear, it's nothing you did. Just seeing how it could get hair quickly and that `egrep` regex I had to use for `BACKGROUNDMET` makes me think this is not going to be a good case for `bash`. It could be jsimilarly structured in Perl.

wwlwpd: Oh, sorry missed a question - right now it's just checking ``$ASGS_CONFIG`.

wwlwpd: After some more thinking, the way to quantify a "linter" is that it checks for things that can be checked without actually executing any code. So the "dumber" the "better". In our case, the linter is really meant to do some high level passes but checking things like a validate date I think is reasonable.  Don't hold me to this, but I think if we can detect something in the ENVIRONMENT+`$ASGS_CONFIG` that will cause `asgs_main.sh` to either fail immediately, or cause a situation that is not discovered until hours later (say during an operational run) _and_ it can be determined _statically_ then we should add it. I have an idea for a self-contained Perl script that will allow us to add things and keep the checks rightly ordered. The checking of the date alone is reason to switch to Perl. So I will whip up a perl based linter and add it to this PR just so we can compare.

jasonfleming: Ok sounds good, thank you for the clarification.

wwlwpd:
```
asgs (SFLV111-nam)> asgs-lint
/ddnA/work/x/asgs/config/mesh_defaults.sh: line 24: allMessage: command not found
(fatal !!!!) CONFIGRATION
config looks NOT OK: /ddnA/work/x/asgs/asgs-configs/cera-daily/SFLv111-nam-supermic.sh
Summary:
  FATAL count: 1
  WARN  count: 0
```

Above is an example of why `source`'ing `$ASGS_CONFIG` is not a good idea:

```
fatalMsg=$(
  source $ASGS_CONFIG
  if [[ ${COLDSTARTDATE:0:4} -lt 1800 || ${COLDSTARTDATE:0:4} -gt 2100 ]]; then
    echo "The COLDSTARTDATE year is outside the expected range of 1800 to 2100."
  fi
  if [[ 10#${COLDSTARTDATE:4:2} -lt 1 || 10#${COLDSTARTDATE:4:2} -gt 12 ]]; then
    echo "The COLDSTARTDATE month is outside the expected range of 1 to 12."
  fi
  # TODO: Improve checking valid days for the given month
  if [[ 10#${COLDSTARTDATE:6:2} -lt 1 || 10#${COLDSTARTDATE:6:2} -gt 31 ]]; then
    echo "The COLDSTARTDATE day is outside the expected range of 1 to 31."
  fi
  if [[ 10#${COLDSTARTDATE:8:2} -lt 0 || 10#${COLDSTARTDATE:8:2} -gt 23 ]]; then
    echo "The COLDSTARTDATE hour is outside the expected range of 0 to 23."
  fi
)
```

The config file I am using, sources 2 other files; these both use a function that is available under `asgs_main.sh` via `messaging/logging.sh` but not in `asgsh`:

1. `config/mesh_defaults.sh`
2. `config/io_defaults.sh`

wwlwpd:
Will add checking ENV/`$ASGS_CONFIG` for:

* `CYCLETIMELIMIT`
* `PPN`
* superficial math check on `NCPU`, `NCPUCAPACITY`, and `NUMWRITERS`

jasonfleming:
Need to think about how this is similar to and different from the `asgsh` `verify` command ... as well as the self-checks that asgs does when it starts up ... how do we distinguish what checks go where

wwlwpd:
`verify` commands actually execute things. all the `perl` scripts are run with `perl -c` which is normally useful as a syntax/load module check, unless you execute code in the `BEGIN {}` block. `python3` does something similar but not identical. The other things are a _netCDF4_ python module test that is actually run; and `adcirc` is also actually run.

jasonfleming:
#360 is also relevant

jasonfleming:
more config issues we can statically check for (mostly pulled from `asgs_main.sh`) : 
1. `BACKGROUNDMET` parameter set to something other than `off` or `namBlend` when `TROPICALCYCLONE` is set to something other than `on`.
2. `BACKGROUNDMET` set to something other than allowable values : `on`, `NAM`, `OWI`, `namBlend`, and `off`.
3. `TRIGGER` must be `ftp`, `rss`, `rssembedded`, or `atcf`.
4. `QUEUESYS` must be `serial`, `mpiexec`, `SLURM`, or `PBS`.
5. `NCPUCAPACITY` must be greater than or at least equal to `NCPU` plus `NUMWRITERS` for each individual scenario.
6. `TROPICALCYCLONE` and `BACKGROUNDMET` cannot both be `on` simultaneously. 
7. `TROPICALCYCLONE` and `BACKGROUNDMET` cannot both be `off` simultaneously. 
8. `LASTSUBDIR` cannot be set to `null` in the config file if the ASGS is trying to hotstart (i.e., `HOTORCOLD=hotstart`), and the `STATEFILE` does not exist.
9. `LASTSUBDIR` cannot be set to `null` in the `STATEFILE` file if it exists and the ASGS is trying to hotstart (i.e.,  `HOTORCOLD=hotstart`).
10. `VORTEXMODEL` parameter set to something other than `SYMMETRIC`, `ASYMMETRIC`, and `GAHM`.
11. `VORTEXMODEL` parameter set to something other than `GAHM` when `BACKGROUNDMET` set to `namBlend`.

jasonfleming:
The startup checks in ASGS look for the files it is supposed to be using (like the mesh file, the stations file etc). My understanding is that this should not go into the linter, because it is not merely a static check of the configuration. But it would be nice if the Operator could manually execute the search for the required files without actually starting ASGS. What do you think?

wwlwpd:
@jasonfleming  - I think a maintainable format for the linter is coming into focus, I still need to implement that great list of "valid" values you posted above (thanks!) but this can definitely be handled.

wwlwpd:
I took care of everything except `VORTEXMODEL` and `QUEUESYS` - wanted to discuss those with you since the former is not commonly seen in configs and the latter is set via `platforms.sh`

jasonfleming:
Yes `VORTEXMODEL` and `QUEUESYS` are probably going to be flagged as config errors rarely. For `QUEUESYS` the only case I can think of is if someone is porting to a new platform and misspells the name of the queueing system in `platforms.sh`. For `VORTEXMODEL`, we don't use anything other than `GAHM` in real time. For hindcasting historical storms (before the age of BEST track files), `SYMMETRIC` is actually the right choice, and we don't support blending for that yet. I can imagine a situation where the `VORTEXMODEL` gets set incorrectly due to copypasta from an existing config file.

wwlwpd:
Okay, sounds like `QUEUESYS` and `VORTEXTMODEL` are set dynamically in most cases. For the linter, they can be _optional_ - as in, "not required, but if set in the config, validate according to the rules setforth..." The other side of the validation/preflight/flight/postflight coin will be a runtime diagnostics component. The linter being "anything that can be checked without running or `source`ing an code".

wwlwpd:
@jasonfleming - I implemented the checks you suggested above. I have a comment in the code, looking for any esoteric or obscure variables that might be easily added (e.g., fort.13, etc).

I was also wondering if there is a relationship between `WAVES` and any of the _SWAN_ stuff we could check; also was wondering if there is something regarding `VARFLUX` that can be checked - basically looking for a brain dump of all the stuff we can check for. So far I have not run into something that, if can be determined _statically_, can't be checked. This includes the existence of files, validation of URIs, remote directories, etc. So let me know.

jasonfleming:
If `WAVES=on` then check the following (if possible with existing framework):

- `$SWANDT` is in a reasonable range (300 to 1200)
- `$SWANDIR` exists
- `swaninit.template` and `$SWANTEMPLATE` files exist in `input/meshes/common/swan` under either `$SCRIPTDIR` or `$ASGS_LOCAL_DIR` (but not both?) and have nonzero length
- `adcswan` and `padcswan` executables exist in `$ADCIRCDIR`
- `unhcat.exe` executable exists in `$SWANDIR`

jasonfleming:
If `VARFLUX=default` then verify that the file `$INPUTDIR/$RIVERFLUX` exists and has nonzero length
if `VARFLUX=on` or `VARFLUX=default` and the ADCIRC version is less than v53, then verify that the file `$INPUTDIR/$RIVERINIT` exists and has nonzero length
If `VARFLUX=on` and `USERIVERFILEONLY=no` then verify that
- `RIVERDATAPROTOCOL` is set to one of `scp`, `ftp`, or `filesystem`
- if `RIVERDATAPROTOCOL=scp` then verify that the ssh server `$RIVERSITE` is reachable as the user `$RIVERUSER` and that the directory `$RIVERDIR` exists and is readable

if `VARFLUX=on` or `VARFLUX=default` for any ADCIRC version,
verify that the file `$INPUTDIR/$HINDCASTRIVERFLUX` exists and has nonzero length

wwlwpd:
> If `VARFLUX=default` then verify that the file `$INPUTDIR/$RIVERFLUX` exists and has nonzero length if `VARFLUX=on` or `VARFLUX=default` and the ADCIRC version is less than v53, then verify that the file `$INPUTDIR/$RIVERINIT` exists and has nonzero length If `VARFLUX=on` and `USERIVERFILEONLY=no` then verify that
> 
>     * `RIVERDATAPROTOCOL` is set to one of `scp`, `ftp`, or `filesystem`
> 
>     * if `RIVERDATAPROTOCOL=scp` then verify that the ssh server `$RIVERSITE` is reachable as the user `$RIVERUSER` and that the directory `$RIVERDIR` exists and is readable
> 
> 
> if `VARFLUX=on` or `VARFLUX=default` for any ADCIRC version, verify that the file `$INPUTDIR/$HINDCASTRIVERFLUX` exists and has nonzero length

@jasonfleming, do you mind clarifying this a bit more for me?

* What's the most independent variable here, i.e., the one all others depend on? (e.g., for SWAN it's `WAVES`)
* Can you tell me more about `$INPUTDIR`? I see it in lots of configs, but is this a river flux boundary specific thing?

wwlwpd:
While the work for the most recent conditions to check are in progress, I wanted to report that I have added a check for _profile_ variables that should never be used in a config (unless you want to have a bad time 😄 )

jasonfleming:
Hey @wwlwpd thank you for working on this ... the independent variable for the river boundary conditions is `VARFLUX`. If can be `on`, `off`, or `default`. If it is set to `default` then the ASGS just applies a static "base flow" condition to the river boundary using a fort.20 file supplied by the Operator. If it is `on`, the ASGS tries to download river flux dynamically from a remote server. However the `VARFLUX=default` capability has not been used for quite a while ... maybe not in the last 5 years, or longer. The team that was running the rainfall runoff model that created the dynamic boundary condition files stopped running their system long ago. 

For `INPUTDIR`, The assumption in ASGS is that the input files (including mesh, fort.15 template, station locations, etc) are found in `INPUTDIR` which is defined in `config/mesh_defaults.sh` and is typically set to something like e.g., `$SCRIPTDIR/input/meshes/HSOFS`.

wwlwpd:
@jasonfleming since the river stuff seems to be tied to getting a value out of `mesh_defaults..sh`, how about round this conclude this round of linter checks without it; This might be better facilitated with any improvements we make as a result of adding site local mesh support - any thoughts?

wwlwpd:
That said, there is a check for `VARFLUX` meeting the set of values: `on`, `off, & `default` - just no subsequent checks on the river values or checking the mesh or input files; also looked like `NCv999` was the only mesh with this set up for river flux out of the box?

wwlwpd:
I added the check for `enableStatusNotify`, treating at important enough to make it required to be set to `yes` or `no`.

jasonfleming:
This looks great, well organized and very extensive. Should I check out this branch and take it for a test drive?

wwlwpd:
I need to add `EMAILNOTIFY` this is what actually effects, `notification.emailnotify : yes`

jasonfleming:
Ok should I review once the `EMAILNOTIFY` check is included?

wwlwpd:
> Ok should I review once the `EMAILNOTIFY` check is included?

Yes, ty :) - will ping you

wwlwpd:
@jasonfleming this is ready to go.

jasonfleming:
May want to provide a heads up that this requires updating `asgsh` (no biggie):
```
asgs (EC95_nam_jgf)> asgs-lint
Can't locate Util/H2O/More.pm in @INC (you may need to install the Util::H2O::More module) (@INC contains: /home/jason/work/opt/default/perl5/perls/perl-5.34.0/lib/site_perl/5.34.0//x86_64-linux /home/jason/work/opt/default/perl5/perls/perl-5.34.0/lib/site_perl/5.34.0/ /home/jason/Campaigns/Development/asgs/PERL /home/jason/work/opt/default/perl5/perls/perl-5.34.0/lib/site_perl/5.34.0/x86_64-linux /home/jason/work/opt/default/perl5/perls/perl-5.34.0/lib/site_perl/5.34.0 /home/jason/work/opt/default/perl5/perls/perl-5.34.0/lib/5.34.0/x86_64-linux /home/jason/work/opt/default/perl5/perls/perl-5.34.0/lib/5.34.0) at /home/jason/Campaigns/Development/asgs/bin/asgs-lint line 8.
BEGIN failed--compilation aborted at /home/jason/Campaigns/Development/asgs/bin/asgs-lint line 8.
```

jasonfleming:
I got this on `ASGSADMIN`:
```
asgs (EC95_nam_jgf)> asgs-lint
(!!! assert_FATAL): 'ASGSADMIN' not defined in '/home/jason/Campaigns/Development/asgs-configs/2022/asgs_config_EC95d_nam_jgf.sh'!
ASGSADMIN - must be an email address
 ... fix and rerun linter
```
I normally keep `ASGSADMIN` set in my `~/.asgsh_profile` on each platform (not every Operator may want this). One nice side benefit of this approach is that my ASGS config file does not have `ASGSADMIN` in it, which makes it portable to other Operators.

jasonfleming:
I got several responses from `asgs-lint` that were not exactly what I expected for the following variables because I normally do not set them in the ASGS config file: `PPN`, `QUEUENAME`, `SERQUEUE`, `TRIGGER`, `NOTIFY_SCRIPT`, `TDS`, and `EMAILNOTIFY`

I normally just use the default values set by `platforms.sh` for `PPN`, `QUEUENAME`, `SERQUEUE`, and `TDS`. Maybe these can only be checked dynamically? 

The default value of `TRIGGER` is set in `config/forcing_defaults.sh`, and the default values of `NOTIFY_SCRIPT` and `EMAILNOTIFY` are set in `variables_init.sh`. Maybe these can only be checked dynamically also?

wwlwpd:
Keep the feedback coming,this is good stuff. I can make most of those "check if in the config", but I do want to keep `EMAILNOTIFY` in there because it's pretty much at the root of this "can't send email" mystery.

jasonfleming:
Ok isn't there an `OPENDAPNOTIFY` also? Does this overlap with `EMAILNOTIFY`?

wwlwpd:
> Ok isn't there an `OPENDAPNOTIFY` also? Does this overlap with `EMAILNOTIFY`?

This defines the emails that get used for sending; `EMAILNOTIFY` controls in the `run.properties` file, `notification.emailnotify`. And if that is "no" (the default in `writeProperties` then no emails get sent in any case with `opdap_post[2].sh`.

`OPENDAPNOTIFY` controls `notification.opendap.email.opendapnotify`.

wwlwpd:
@jasonfleming made all the ones you mentioned "option but validated if defined" except for `EMAILNOTIFY` - we really need to enforce this as long as we're in the current situation we are with notifications, etc.

Once this is put into master, we can create a new issue to tweak based on feedback. At some point, probably after this season; we should feel comfortable enough to run `asgs-lint` as part of the `asgs_main.sh` start up process. We can also start thinking about how to incorporate dynamic checks - which I agree includes many of the environmental variables.

jasonfleming:
Ok I added `EMAILNOTIFY` to my config file and gave it another shot. I got the following:
```
asgs (EC95_nam_jgf)> asgs-lint
(!!! assert_FATAL): ADCIRCDIR and ADCIRC_BINS must be available in the environment.
 ... fix and rerun linter
asgs (EC95_nam_jgf)> echo $ADCIRCDIR
/home/jason/work/adcirc-cg-v53release-gfortran/work
asgs (EC95_nam_jgf)> echo $ADCIRC_BINS

asgs (EC95_nam_jgf)> 
```
It looks like I had built ADCIRC manually. I will try rebuilding with `build adcirc` to see if that fixes it.

wwlwpd:
For a manual build, replicate an ADCIRC "profile" in `~/.adcirc-meta` - that's the way to work around what's done by `build adcirc` and what is exected to be in the environment once you load something. I'd add something like `-manual`. I can see us being able to automatically "registered" an ADCIRC build and generate this file automatically in the future. I'll add a GH issue for that.

jasonfleming:
Ok I was able to use `build adcirc` on my desktop machine after fixing a typo in `cmplrflags.mk`. I will issue a separate PR for that one. :-)

jasonfleming (approved):
This looks great @wwlwpd thank you for all your effort on this, it will continue to pay dividends as we continue to add all manner of static checks to this framework in the future!

wwlwpd:
Thank you, too - this has been kind of a slog but not that bad. I think this will be a great first iteration and am excited to see what else we can do with the tool.

PR #786, opened 2022-03-25
adding support for local directory for site-specific assets

jasonfleming:
Added support for `local/platforms.sh` and `local/config/mesh_defaults.sh`. Still need to add `local/bin` to `$PATH`, an environmental variable to place `local` in some other directory, etc.

### DIFF
--- a/config/mesh_defaults.sh
+++ b/config/mesh_defaults.sh
@@ -742,6 +742,13 @@ case $GRIDNAME in
       UNITOFFSETFILE=null
       ;;
    *)
-      warn "cycle $CYCLE: $SCENARIO: $THIS: Mesh GRIDNAME $GRIDNAME was not recognized."
+      localMeshParameters=$SCRIPTDIR/local/config/mesh_defaults.sh
+      logMessage "cycle $CYCLE: $SCENARIO: $THIS: Mesh GRIDNAME $GRIDNAME was not found in config/mesh_defaults.sh so now ASGS is checking the file $localMeshParameters (if it exists)."
+      if [[ -e $localMeshParameters ]]; then
+         logMessage "cycle $CYCLE: $SCENARIO: $THIS: $localMeshParameters was found."
+         source $localMeshParameters
+      else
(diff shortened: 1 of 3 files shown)


platforms.sh
```diff
@@ -827,3 +827,11 @@ HPC_Reservation_Hint()
    esac
 }
 
+localPlatformParameters=$SCRIPTDIR/local/platforms.sh
```
wwlwpd, line 830: I would wait to do the platforms.sh, since it has some things in there that are pretty specific and idk yet how we'd implement those generally (`HPCENV`, the _Hints_, etc)

config/mesh_defaults.sh
```diff
@@ -742,6 +742,13 @@ case $GRIDNAME in
       UNITOFFSETFILE=null
       ;;
    *)
-      warn "cycle $CYCLE: $SCENARIO: $THIS: Mesh GRIDNAME $GRIDNAME was not recognized."
+      localMeshParameters=$SCRIPTDIR/local/config/mesh_defaults.sh
```
wwlwpd, line 745: Was this going to include a configurable base directory envar? We'll want to make sure we add this to be tracked by the profiles; the list of envars the shell will save is tracked in `_ASGS_EXPORTED_VARS_` so any "local" base dir variable name should be added there, see https://github.com/StormSurgeLive/asgs/blob/master/cloud/general/asgs-brew.pl#L452

This way, setting a "local" director per profile is then possible.

wwlwpd (commented):
Added some preliminary comments. I know this is a WIP.

platforms.sh
jasonfleming: Ok I will comment that out.

config/mesh_defaults.sh
jasonfleming: I'm not confident that I can do this without further guidance ... let's discuss.

jasonfleming:
Let's hold off on this until the `platforms` PR #798 is merged.

wwlwpd:
Yes let's discuss. I am going to close this for now.

PR #787, opened 2022-03-28
Changed default ASGS log dir to $WORK/log

### LINKED ISSUE #51: make location of asgs log file configurable in asgs config file
By default, the ASGS log file is written to the same directory where the ASGS is started. This is opaque somewhat non-intuitive. If the log file is stored in the Operator's home directory, lengthy and/or numerous asgs log files can fill up the Operator's home directory. We should have a new ASGS variable called SYSLOGDIR. We should default to setting SYSLOGDIR=$SCRATCHDIR/asgs/log (making the directory if necessary) and give the Operator the ability to set SYSLOGDIR in each asgs config file.

jasonfleming:
Resolves #51 ... This is a change in behavior ... previously the ASGS log file `$SYSLOG` was simply dropped in whatever  `$PWD` was set to when `asgs_main.sh` was executed, which was suboptimal, with log files cluttering up my `$HOME`, `$SCRIPTDIR` etc. Now, the default directory for the ASGS log file `$SYSLOG` is set to `$WORK/log` upon startup (creating that directory if necessary). Operators can change $SYSLOG in their ASGS config files if they want to.

### DIFF
--- a/monitoring/logging.sh
+++ b/monitoring/logging.sh
@@ -251,7 +251,15 @@ RMQMessageRunProp()
 # set the name of the asgs log file
 setSyslogFileName()
 {
-   SYSLOG=`pwd`/${INSTANCENAME}.asgs-${STARTDATETIME}.$$.log
+   local DATETIME=$(date +'%Y-%h-%d-T%H:%M:%S%z')
+   if [[ ! -d $WORK/log ]]; then
+      mkdir -p $WORK/log
+      echo "[$DATETIME] INFO: Created subdirectory for ASGS log files : '$WORK/log'."
+   else
+      echo "[$DATETIME] INFO: Found subdirectory for ASGS log files : '$WORK/log'."
+   fi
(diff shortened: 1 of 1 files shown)


monitoring/logging.sh
```diff
@@ -251,7 +251,15 @@ RMQMessageRunProp()
 # set the name of the asgs log file
 setSyslogFileName()
 {
-   SYSLOG=`pwd`/${INSTANCENAME}.asgs-${STARTDATETIME}.$$.log
+   local DATETIME=$(date +'%Y-%h-%d-T%H:%M:%S%z')
+   if [[ ! -d $WORK/log ]]; then
+      mkdir -p $WORK/log
```
wwlwpd, line 256: This all looks like a good change, so this is just a note. A lot of times I don't find it necessary to check to see if a directory is created if I am going to create it directly, so I tend to do something like this:

```
# mkdir without checking,
mkdir -p $WORK/log 2> /dev/null
```
However this doesn't cover the cases of:

1. `$WORK/log` exists but is a file
2. `mkdir -p` fails for a reason related to a filesystem issue

wwlwpd (approved):

PR #792, opened 2022-03-30
made NSCREEN configurable via ASGS config file

### LINKED ISSUE #780: templatize or standardize `NSCREEN` value in all fort.15 in supported meshes
@jasonfleming - it'd be supremely helpful if we standardized this to be a reasonable negative number, so it generated reliably an `adcirc.log` (per ADCIRC documentation) that provides stepwise output the current model situation - I've ALWAYS found this output valuable, though I just now realized if <0 it'll write to a `adcirc.log`.

You can see the spread of values (but not the actual distribution) with, 

```
find ./input/meshes -type f | xargs grep -h NSCREEN | awk '{print $1}' | sort -nr | uniq
```

It's all over the place,

```
2000
1000
600
100
1
-1
-100
-1000
-2000
```
For this issue, I'd like to see what you thought about me standardizing on, say `-10` as a template variable default? I'd update all the fort.15s in all the meshes currently supported.

As future work, it would pretty simple to design some "monitoring" that can check this file by parsing the latest line and also maybe by tracking how often it is completing "10" timesteps for a fairly good sample  to compute module throughput (e.g., X ts/minute). The output line also has other valuable information, as you know, timestep, min/max values for elevation and velocities as well as "%" complete (wrt to total `RNDAY`, `DT`, and current timestep). But we can't do anything reliably if `NSCREEN` is inconsistently applied everywhere.

This would also help greatly with debugging and questions like, _is it really running?_

jasonfleming:
resolves #780 

Gave this a reasonable default value of -1000


config/io_defaults.sh
```diff
@@ -23,6 +23,7 @@
 THIS=$(basename -- $0)
 allMessage "$THIS: Setting default parameters for input/output."
 #
+NSCREEN=-1000 # frequency (in time steps) of output to adcirc.log
```
wwlwpd, line 26: If we want to be able to set this in a configuration file; then I think this needs to be something like,

```
NSCREEN=${NSCREEN:-"-1000"}
```

Do you thing?

wwlwpd (changes requested):
Final comment - do you have some changes to `./asgs_main.sh` and `./control_file_gen.pl` that you forgot to add to this commit, or is that not needed?

output/inlinec-hw.pl
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env perl
```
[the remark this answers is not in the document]
jasonfleming: No, I will delete it, good catch.

jasonfleming:
I have added the `control_file_gen.pl` and `asgs_main.sh` changes. I guess I had a mental block against making more spaghetti. :-) I will test these changes on my desktop machine to verify that they work as expected and then send another update.

control_file_gen.pl
```diff
@@ -70,7 +70,7 @@
 use Date::Calc;
 use Cwd;
 #
-
+my $nscreen=1;    # frequency of time step output to STDOUT(+) or adcirc.log(-)
```
wwlwpd, line 73: should this match the default, `-1000` ?

control_file_gen.pl
```diff
@@ -175,6 +175,7 @@
            "hstime=s" => \$hstime,
            "advisdir=s" => \$advisdir,
            "scriptdir=s" => \$scriptdir,
+           "nscreen=s" => \$nscreen,
```
wwlwpd, line 178: Super tiny _nit_ but can you align the `=>` with the one above or below?

asgs_main.sh
```diff
@@ -3171,7 +3171,7 @@ while [ true ]; do
       CONTROLOPTIONS="$CONTROLOPTIONS --periodicflux $PERIODICFLUX"  # for specifying constant periodic flux
       RMQMessage "INFO" "$CURRENT_EVENT" "$THIS>$ENSTORM" "$CURRENT_STATE" "Generating ADCIRC Control File (fort.15) for $ENSTORM."
       logMessage "$ENSTORM: $THIS: Generating ADCIRC Control File (fort.15) for $ENSTORM with the following options: $CONTROLOPTIONS."
-
+      CONTROLOPTIONS="$CONTROLOPTIONS --nscreen $NSCREEN"
```
wwlwpd, line 3174: I had a question - will the default for `$NSCREEN` always be available via `io_defaults.sh` or should this default value of `-1000` actually reside in `asgs_main.sh`?

wwlwpd (changes requested):
Had some questions, comments - biggest thing I saw was that the default value in `control_file_gen.pl` didn't match the other default of `-1000`.

control_file_gen.pl
jasonfleming: I did it this way so that we would know immediately if there was an issue where the `NSCREEN` was not populating correctly from the default value in `io_defaults.sh`. I'll change it so the defaults match.

control_file_gen.pl
jasonfleming: of course

asgs_main.sh
jasonfleming: Ok I have added this to `variables_init()`

output/inlinec-hw.pl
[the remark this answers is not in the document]
jasonfleming: Removed.

config/io_defaults.sh
jasonfleming: Yes, I have now set it that way in the code in this PR.

control_file_gen.pl
jasonfleming: Code for `control_file_gen.pl` in the PR now reflects consistent default value of -1000.

control_file_gen.pl
jasonfleming: Fat arrow now lines up with the one above.

wwlwpd (approved):
Looks good to me!

PR #797, opened 2022-03-31
Added --insecure flag that could be useful for storm testing.

### LINKED ISSUE #796: add `--insecure` flag to `get_atcf.pl`
Refinement to the control over `https` (default) versus `http`.

wwlwpd:
Issue 796: StormReplay.com now supports https, so there was no need
to restrict https only to NOAA NHC assets. This change adds a flag
that maybe set when the RSS WWW site needs to use 'http' over 'https'.

Resolves #796.


wwlwpd:
Merging, simple change; low risk.

PR #800, opened 2022-04-12
Set default NSCREEN to -1000 and added %NSCREEN% to all fort.15 templ…

### LINKED ISSUE #780: templatize or standardize `NSCREEN` value in all fort.15 in supported meshes

wwlwpd:
…ate files

Issue 780: Added NSCREEN to parameters list for control_file_gen.pl
as passed from asgs_main.sh; removed stray file that shoud not
have been committed Modified setting of default NSCREEN so it left
Operator configured value in place Added NCSREEN to variables_init()
and reconciled its default values.

Resolves #780.

Supersedes PR #792.


wwlwpd:
Disregard

Issue #794, opened 2022-03-30
moving over an existing profile to another ASGS on the same machine is cumbersome


wwlwpd:
If you're using a profile created on one asgs installation on a second one, `SCRIPTDIR` will need to be updated; so right now what I recommend is starting the new `./asgsh` ; then:

1. `list profile` - note the name of the "default" profile for the asgs you're in
2. `load profile`  - select the "old" profile you want to use in the new installation
3. `clone profile` - this will create the new profile you'll be using (name it something meaningful but different)
4. `show config` - copy the path to the current $ASGS_CONFIG
5. `rebuild profile`  - see below...

For #5, you will want:

- base the profile off of the new default, noted in #1 above
- provide the past to the config, noted in #4 above

To fix this, a few things may be in order:

* track the original _default_ profile that's created on _./init-asgs.sh_
* automate `clone profile` or `rebuild profile`, maybe provide a command that does both with some assumptions made (minimal questions asked)

jasonfleming:
Maybe we could add an `import profile` command that would automate 2, 3, and 5 above? 

Maybe we could also keep track of `SCRIPTDIR` as well as the currently checked out branch via the `asgsh` prompt? In other words, if the `SCRIPTDIR` is named asgs and the master branch is currently checked out, the `asgsh` prompt with the default profile could be something like
```
asgs:master (default)>
```
 As another example, if the `SCRIPTDIR` is named `staging`, the `issue-999` branch is checked out, and the profile is `HSOFS_nam_op` then the prompt might be
```
staging:issue-999 (HSOFS_nam_op)> 
```
What do you think?

wwlwpd:
I will need to rethink this in lieu of #880 - since in an isolated ASGS, `clone profile` is a lot easier and there are no concerns anymore about "portability".

wwlwpd:
`import profile` could be useful - I won't close this yet and we can think about it. TY for your feedback above.

PR #807, opened 2022-04-20
Remove SCRIPTDIR from save profiles, set in local ./asgsh.

### LINKED ISSUE #806: Remove SCRIPTDIR from save profiles, set in local ./asgsh
related to #794, but doesn't directly address that step.

wwlwpd:
Issue 806: Isolated ASGS should share profiles, but SCRIPTDIR is a
residual setting in the profile. It probably should be set by the
local ./asgsh and not included in the profile.

If there are any issues with this change, they should only creep
up for users who are leveraging multiple ASGS installations on
the same user account, same machine.

Additionally,
  * ./update-asgs --clean (passed to asgs-brew.pl) now leaves
    existing ASGS profiles and existing ADCIRC profiles in the
    directories $HOME/.asgs and $HOME/.adcirc-meta (respectively).

  * ASGS_META_DIR replaces instances of "$HOME/.asgs", which may
    be used in theory to relocate profile locations (in the future)

Related to #794 by making profiles more independent of specific
ASGS installations.

Resolves #806.


jasonfleming (approved):
This looks good, thank you. Its important to find and resolve these subtle issues raised by the self-containment of multiple ASGS versions/branches on one platform. I like the `ASGS_META_DIR` varname, it does just what it sounds like it does. 

Let us know if there is a "breaking" change, i.e., something that would require Operators to need to reinstall from scratch (which is easy to do).

PR #812, opened 2022-04-26
invasive correction for mail integration

wwlwpd:
* moved $SCRIPTDIR/asgs-sendmail.pl -> $SCRIPTDIR/bin/asgs-sendmail (so it's in `asgsh`'s `$PATH`)
* updated all places in repo that used it, except one place that was attaching files
* NOTE: some places were still using the system `mail` utility, which was unexpected (my fault)

### DIFF
--- a/bin/asgs-sendmail
+++ b/bin/asgs-sendmail
@@ -1,6 +1,6 @@
 #!/usr/bin/env perl
 #--------------------------------------------------------------------------
-# asgs-sendmail.pl
+# asgs-sendmail
 #--------------------------------------------------------------------------
 # Copyright(C) 2019 Brett Estrade
 # Copyright(C) 2019 Jason Fleming
@@ -20,7 +20,7 @@
 # You should have received a copy of the GNU General Public License
 # along with the ASGS.  If not, see <http://www.gnu.org/licenses/>.
 #--------------------------------------------------------------------------
-package util::asgs_sendmail;
+package bin::asgs_sendmail;
 
 use strict;
 use warnings;
--- a/init-asgs.sh
+++ b/init-asgs.sh
@@ -44,7 +44,7 @@ while getopts "bmp:x:" optname; do
            exit 1
          fi
          ;;
-      p) export ASGS_LOCAL_DIR=${OPTARG}
+      p) export ASGS_LOCAL_DIR=$(readlink -f "${OPTARG}") # get full path
          ;;
       x) # add extra arbitrary options to asgs-brew.pl command
          if [ -z "${EXTRA_ASGSBREW_OPTS}" ]; then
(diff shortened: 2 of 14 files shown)


wwlwpd:
@jasonfleming when you review this, make sure `asgs-sendmail` usage fits the following criteria:

* assume it's available via `$PATH`
* email body should be getting redirected INTO the script
* flags `--subject` and `--to` should be the only ones used
* shell variables being used for `--subject` and `--to` should be contained in DOUBLE qoutes

Examples,

* `cat somefile.txt | asgs-sendmail --subject "some subject $OTHERSUBJECTINFO" --to "$SOME_EMAIL_LIST" 2>> $SYSLOG`
* `asgs-sendmail --subject "some subject $OTHERSUBJECTINFO" --to "$SOME_EMAIL_LIST" < somefile.txt 2>> $SYSLOG`

Let me know if something doesn't look right. `asgs-sendmail` is not used where attachments are added (only one place I think).

jasonfleming (changes requested):
It may not matter but line 401 in `opendap_post.sh` does not have `""` around `$OPENDAPNOTIFY` for the `--to` cmd line arg

Lines 421, 535, 605, and 638 in `opendap_post2.sh` do not enclose `$OPENDAPNOTIFY` in `""` for the `--to` cmd line arg

line 248 in `scp_only.sh` does not enclose `$OPENDAPNOTIFY` in `""` for the `--to` cmd line arg

Is it possible to check `$?` after `asgs-sendmail` is executed and issue a `warn()` message if it does not come back with `0`?

wwlwpd:
great thank you, I can do all that - I want to delete `opendap_post.sh` or rename `opendap_post2.sh` to that name..thoughts?

jasonfleming:
Delete/rename: That is fine, but we have a good system now where the script name can be arbitrary b/c it is a shell variable rather than being hard coded. 

I think you mentioned at some point wanting to get rid of the term "opendap" b/c it is not actually relevant to the action performed by this script. I agree. Can we rename this script `upload.sh` and make that name the default?

wwlwpd:
> Delete/rename: That is fine, but we have a good system now where the script name can be arbitrary b/c it is a shell variable rather than being hard coded.
> 
> I think you mentioned at some point wanting to get rid of the term "opendap" b/c it is not actually relevant to the action performed by this script. I agree. Can we rename this script `upload.sh` and make that name the default?

I am not against that, but despite my proactive tone at the start I think this would just confuse things :) - we can do it some other time, though.

wwlwpd:
Changes request were made, this is ready for your review. I think you have to "resolve conversation" since you requested the change in the review.

wwlwpd:
but do take some time with this one, not trying to add pressure (none that's not already there :))

jasonfleming (approved):

jasonfleming:
Great to have a simpler script to work with. Name the script as you see fit, I am open to anything. Please go ahead and merge. :-)

PR #823, opened 2022-05-02
If ASGS_LOCAL_DIR is defined, asgs-brew.pl will add it to PATH.

### LINKED ISSUE #821: ASGS should add `$ASGS_LOCAL_DIR/bin` to `PATH` for local site support
related to #790

wwlwpd:
Issue 821: When ASGS_LOCAL_DIR is specified during ./init-asgs.sh's
"-p" option, the asgs-brew.pl step that does the installation and
environmental set up will now add $ASGS_LOCAL_DIR/bin to PATH if
$ASGS_LOCAL_DIR/bin does actually exist.

The net result is that executables contained in $ASGS_LOCAL_DIR/bin
will be available via PATH, which offers parity with how the env
treats $SCRIPTDIR/bin.

Resolves #821.


jasonfleming (approved):
Can we add `$ASGS_LOCAL_DIR/bin` whether or not it exists at the time that `asgs-brew.pl` is run? I'm thinking of a situation where that directory does not exist at installation time, but is then added later by the Operator but then `asgsh` can't find it without running `asgs-brew.pl` again. In any case, this looks great, thank you.

wwlwpd:
Great idea! Will do.

wwlwpd:
@jasonfleming made the addition to `PATH` less restrictive as you suggested; ready for another look

PR #839, opened 2022-06-04
Tweaks to get running on lonestar6 (TACC)

wwlwpd:


jasonfleming (dismissed):
@wwlwpd this looks great, thank you. Do I undersdand correctly that the files are now `scp`'d to the VM at TACC and that is how they get to Corral? If so, this is the reverse of the way we were doing things before, and more consistent with other platforms and HPC centers. The only thing I am wondering about is the unix group that these files take on once they get there. Also, I wonder what would happen if we exceed our Corral quota ... will the `scp` just fail? Will we get an error message that says disk quota exceeded? In any case, this looks great, thank you.

wwlwpd:
> Do I undersdand correctly that the files are now scp'd to the VM at TACC and that is how they get to Corral?

@jasonfleming funny enough, it just worked once I updated the the `OPENDAPBASEDIR` to reflect the mounted director on the virtual machine hosting THREDDS; but moreover I didn't even think about not `scp`ing them over - so I guess the answer to your first question is, YES. It works like the rest of them.

I do not know about the quota, but I suspect this would be enforced via the NFS mount on the virtual machine and by virtue of the effective user used for `scp` access. So ASGS is treating the remote server as a blackbox like all the others.

Note: I created a "new" way to add remote servers from now on which is required to support the eventual goal for supporting "local" remote servers over `scp`/`ssh` (#786).

jasonfleming:
Yeah I saw that new way to add remote servers. Very nice.

wwlwpd:
I will merge this, you may switch to using `master` after - any issues we find after we'll just fix under new GH issues.

wwlwpd:
GH knocked off the approval by @jasonfleming, but this is the same change. Just updated the commit messages and rebased on `master`.

PR #842, opened 2022-06-09
Adding `HINDCASTONCE_AND_EXIT` for running hindcast once, then exiting.

wwlwpd:
Issue 841:

WIP.

### DIFF
--- a/asgs_main.sh
+++ b/asgs_main.sh
@@ -2108,6 +2108,14 @@ CYCLELOG=null
 SCENARIOLOG=null
 #
 executeHookScripts "FINISH_SPINUP_STAGE"
+
+# Use for hindcast activities to end ASGS after the hindcast has been run
+if [[ "${HINDCASTONCE_AND_EXIT}" == "y" || "${HINDCASTONCE_AND_EXIT}" == "yes" ]]; then
+   logMessage "$THIS: Shutting down due to HINDCASTONCE_AND_EXIT mode ..."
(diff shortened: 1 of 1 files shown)


jasonfleming (approved):
This is cool, go for it. Also, we do have the `ONESHOT` parameter in `asgs_main.sh`, which is set to `yes` when the `-s $STATEFILE` option is specified on the command line to `asgs_main.sh.` If `$ONESHOT == "yes"` then `asgs_main.sh` will `exit` at the end of the forecast stage, rather than going back to look for another nowcast. This may be what we want here (or not depending on the specific behavior you are interested in). I added this `ONESHOT` feature when we had an interest in running ASGS in the NOAA Ops center (in their world, everything has to run as a `cron` job). So the ASGS couldn't be a persistent process, and had to be awakened by `cron` periodically and be given a file with its state information so it would know where it left off. If it had something to do (i.e., a new forecast cycle) it would do it and then exit, leaving behind an updated statefile.

oodler577:
I am going to run it with `ONESHOT=yes` in the config, that may be sufficient. Thanks!

wwlwpd:
It looks like `ONESHOT` internally does a couple of things that are not really desired. E.g., it doesn't generate a `STATEFILE`; I think that `HINDCASTONCE_AND_EXIT` is a good addition and that `ONESHOT` is useful for a different situation, like what you had described.

PR #852, opened 2022-06-15
Removed erroneous check for the existence of the fort.26 file

jasonfleming:
resolves #849


jasonfleming:
This was a cut-and-paste error. There should not be a check for the `fort.26` file during the so-called "hindcast" phase. There is no meteorological forcing during the tidal/river initialization, so under no circumstances would there be a `fort.26` file. The enhanced checking for this was cut and pasted from the nowcast and forecast phases erroneously.

wwlwpd (approved):
Looks great, issue with the failure mode has been resolved.

PR #853, opened 2022-06-16
Ensure STORM is zero-padded for single number storms.

### LINKED ISSUE #827: remove requirement that single digit storm number have a leading zero in asgs config file
If an Operator sets the tropical cyclone storm number to a single digit in the ASGS config file, e.g. `STORM=9`, the ASGS will have issues because it uses this number as a string in various file names etc. This should be resolved so that Operators can use `STORM=09` or `STORM=9` in their config files and not have to worry about it. Either way, the ASGS should look for files named e.g., `bal092008.dat` etc.

wwlwpd:
Issue 827: This change conditions $STORM setting by zero padding
single numbers set in a configuration file; this is a common mistake
that happens during operations, and zero padding is required for
scripts used through the operation forecast runs. It is also required
for properly constructing the full storm designations issued by
the NHC.

Resolves #827.

### DIFF
--- a/asgs_main.sh
+++ b/asgs_main.sh
@@ -53,6 +53,10 @@ readConfig()
    source ${SCRIPTDIR}/config/forcing_defaults.sh
    # pick up config parameters, set by the Operator, that differ from the defaults
    source ${CONFIG}
+   # ensure single digit STORM numbers issued by NHC are zero-padded
+   if [ -n "${STORM}" ]; then
+     STORM=$(printf "%02d" "$STORM")
+   fi
    # maintain backward compatibility with old config files
    if [[ $ENSEMBLESIZE != "null" ]]; then
        SCENARIOPACKAGESIZE=$ENSEMBLESIZE


jasonfleming (commented):
I think I attempted this in the past, but then had issues where `STORM` was being used as a number in certain places, and the leading zero made bash think the number was octal. I think. I may not be remembering that right, or maybe it was the advisory number. In any case, let's test this with a `STORM` less than 10 to see if it hiccups anywhere. What do you think?

wwlwpd:
Oh I see. We can try that. We can also solve the issue inside of the Perl script(s) that might use it. Thanks!

wwlwpd:
converting this to draft

wwlwpd:
@jasonfleming this seems to be working fine for me, and I've looked through and see no reason why it wouldn't be doing exactly what we want. Can you give it another look?

jasonfleming (approved):
Thank you Brett, this looks good. Another good way to take something of the Operators minds. :-)

PR #866, opened 2022-06-28
Issue 865

### LINKED ISSUE #865: bin/guess platform only works when executed when $PWD is $SCRIPTDIR
The `bin/guess platform` command only works when executed in `$SCRIPTDIR`:
```
asgs (default-asgs)> cd $SCRATCH
asgs (default-asgs)> guess platform
asgs (default-asgs)> cd $SCRIPTDIR
asgs (default-asgs)> guess platform
ls6
```
This appears to be because the `find` command in `bin/guess` used to locate `init.sh` starts in `.` so it only works correctly if the `run` command is issued by the Operator when in `$SCRIPTDIR`. I'm working on a fix to make the search start in `$SCRIPTDIR`.

jasonfleming:
resolves #865 

There was also an issue where $PLATFORM_INIT was not getting set so it was not being sourced to pull in parameters. I've added a workaround for this, but I think it needs to be discussed, because I wasn't sure whether PLATFORM_INIT is supposed to be in the environment from the installation process.

### DIFF
--- a/asgs_main.sh
+++ b/asgs_main.sh
@@ -1528,6 +1528,7 @@ STARTDATETIME=`date +'%Y-%h-%d-T%H:%M:%S%z'`
 SCRIPTDIR=${0%%/asgs_main.sh}  # ASGS scripts/executables
 
 variables_init           # initialize variables
+HPCENVSHORT="null"       # in case this is leftover in the env
 initFileStatusMonitoring # initialize variables
 #
 # create directories with default permissions of "775" and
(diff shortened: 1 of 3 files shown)


wwlwpd:
I thought I had fixed this. Let me take a look at the fix and the error messages and I'll let you know.

wwlwpd (commented):
In principle I approve of these changes, I want to squash them and format the code a little better; which I will take care of.

wwlwpd:
@jasonfleming ping me when you have updated this like you mentioned

wwlwpd:
Now that I am looking at this, `PLATFORM_INIT` should be getting set when `./asgsh` is generated via `asgs-brew.pl` - it doesn't belong in `./platforms.sh`. However, the issue with `bin/guess` is legitimate. Still looking.

wwlwpd:
Closing this in lieu of a different branch, sorry for the confusion, @jasonfleming

PR #878, opened 2022-06-30
Fix bin/guess not being relocatable outside of $SCRIPTDIR

### LINKED ISSUE #865: bin/guess platform only works when executed when $PWD is $SCRIPTDIR

wwlwpd:
Issue 865: Fix must work inside of asgsh in all cases; and
when not in asgsh (SCRIPTDIR is not set), so default to "./",
since it is needed on for ./init-asgs.sh in that case.

Resolves #865.

### DIFF
--- a/bin/guess
+++ b/bin/guess
@@ -40,14 +40,15 @@ _guess_platform()
   fi
 
   # IF NOT FOUND ABOVE ... dive into platforms defined in ./platforms/
-  _PLATFORMS=()
+  local _PLATFORMS=()
+  local _BASE=${SCRIPTDIR:-.} # needs to operate outside of asgsh, for ./init-asgs.sh
   if [ $default_platform == unknown ]; then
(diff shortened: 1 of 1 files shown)


jasonfleming (approved):
This looks great thank you :-)

wwlwpd:
thanks for bearing with me here...

PR #889, opened 2022-07-07
Made best track reading more robust.

### LINKED ISSUE #888: make `./storm_track_gen.pl` reading of best track data more robust

wwlwpd:
Issue 888: Changes made while proving out replay server. If there
are issues during a real storm, it might be related to this commit.

Resolves #888.

### DIFF
--- a/storm_track_gen.pl
+++ b/storm_track_gen.pl
@@ -240,18 +240,18 @@ if ( $match == 0 && $percent ne "null" ) {
 #
 # open ATCF input files
 my $forecastATCF = "$dir/al$storm$year.fst";
-unless (open(FCST,"<$forecastATCF")) {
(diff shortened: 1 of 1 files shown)


wwlwpd:
Current files served by replay server work without these changes, but these changes will make the process more robust - reading treats the "fixed" width columns as fungible but the writing still treats things as fixed width, so there is no loss in the format of any files written; only a little play on the format of the files read (which I think makes it more robust). Hopefully it does.

jasonfleming (approved):
This looks good to me, were the files that needed a bit of "play" in the parser come from NHC or were they generated by `nhc_advisory_bot.pl`? If that script is churning out incorrectly formatted files, it would be good to flag that and see about a fix there. I wonder if the linter can be extended to look at these files and verify correct formatting.

PR #892, opened 2022-07-08
Added --nscreen that seemed to be missing.

### LINKED ISSUE #891: `NSCREEN` set in ASGS configuration file only applied to `fort.15` _after_ the hindcast phase
Not sure if this is by design, but something to note and probably fix.

wwlwpd:
Issue 891:

Resolves #891.


jasonfleming (approved):
This was an oversight on my part. Thank you for finding and fixing.

wwlwpd:
Thanks! I reviewed the original `NSCREEN` support change, so I am to blame as well :)

PR #893, opened 2022-07-08
Checking ownership before changing permissions and redirecting log messages

### LINKED ISSUE #874: Reduce chattiness and checking remote permissions in opendap_post2.sh
When posting to most remote servers (not fortytwo), the opendap_post2.sh script must check to see if it is creating directories, and if so, will most likely need to change permissions on those directories so that other Operators can also post there. However, it is often the case that it is not the entire directory hierarchy that can be changed (some directories are not owned by the Operator doing the posting). This causes the attempt to change permissions in those directories to fail and then retry (up to 10 times each time). This results in slow posting and huge log files that are very hard to parse or post.

On remote servers where permissions are an issue, the attempt to change permissions should be limited to cases where it is actually needed and will succeeed.

jasonfleming:
resolves #874

### DIFF
--- a/output/opendap_post2.sh
+++ b/output/opendap_post2.sh
@@ -21,6 +21,7 @@
 #------------------------------------------------------------------------
 #
 THIS=$(basename -- $0)
+APPLOG=$THIS.log
 #
 EXIT_SUCCESS=0
 EXIT_ERROR=1
@@ -129,8 +130,8 @@ for server in ${SERVERS[*]}; do
    if [[ $SCENARIO == "asgs.instance.status" ]]; then
       cat run.properties >> $RUNPROPERTIES
       rm run.properties # so we don't keep appending to it
-      $SCRIPTDIR/metadata.pl --redact --jsonify --metadatafile $RUNPROPERTIES --converted-file-name asgs.instance.status.json 2>> $SYSLOG
-      sed --in-place "s/$USER/\$USER/g" asgs.instance.status.json 2>> $SYSLOG
+      $SCRIPTDIR/metadata.pl --redact --jsonify --metadatafile $RUNPROPERTIES --converted-file-name asgs.instance.status.json 2>> $APPLOG
+      sed --in-place "s/$USER/\$USER/g" asgs.instance.status.json 2>> $APPLOG
    fi
    # FIXME: enable Operator to override TDS parameter settings from platforms.sh
    _THIS="output/opendap_post2.sh-->$server"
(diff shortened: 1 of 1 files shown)


jasonfleming:
I am testing this now.

output/opendap_post2.sh
```diff
@@ -811,9 +818,9 @@ SSHCMD
       threddsPostStatus=fail
       MSGS="$SCENARIO: $_THIS: The opendap post method $OPENDAPPOSTMETHOD was not recognized."
       if [ "$MANUAL" == 1 ]; then
-        echo "$MSG"
+         echo "$MSG"
```
wwlwpd, line 821: seems like a leading white space was added, other lines have this above also

output/opendap_post2.sh
```diff
@@ -21,6 +21,7 @@
 #------------------------------------------------------------------------
 #
 THIS=$(basename -- $0)
+APPLOG=$THIS.log
```
wwlwpd, line 24: Can you explain what changes are made here and what `$APPLOG` is meant to be? I am definitely not against better logging, but is this creating an new log file type? E.g., will support for something like `tailf applog` be needed.

wwlwpd (changes requested):
Had some questions and minor formatting nits.

output/opendap_post2.sh
jasonfleming: Yes I am diverting the log messages from this script into their own file instead of sending them to the main log file. There are a lot of them and they seem to be filling up that file and making it difficult to read. The messages are really only useful for troubleshooting so it seems that they could be checked manually (rather than using something like `tailf opendap_post.sh.log`)

output/opendap_post2.sh
```diff
@@ -435,23 +438,25 @@ SSHCMD
...
+               echo "$MSG" >> "$APPLOG"
             fi
             unset MSG
-            ssh $OPENDAPHOST bash <<SSHCMD >> $SYSLOG 2>&1
+            ssh $OPENDAPHOST bash <<SSHCMD >> $APPLOG 2>&1
 # this block will be executed on the remote server,
 # variables are interpolated locally unless escaped
 # with a backslash, '\'
-ln -s "$OPENDAPBASEDIR/$STORMNAMEPATH" "$OPENDAPBASEDIR/$ALTSTORMNAMEPATH"
+if [[ ! -e "$OPENDAPBASEDIR/$ALTSTORMNAMEPATH" ]]; then
```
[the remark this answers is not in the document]
jasonfleming: Ok I have added this and will commit/push.

output/opendap_post2.sh
jasonfleming: This script has a couple different indentation schemes going on, I will revert and commit/push.

output/opendap_post2.sh
wwlwpd: @jasonfleming that works for me!

PR #898, opened 2022-07-09
Removed resetting SCRIPTDIR to pwd

### LINKED ISSUE #864: Reloading profile with rl resets SCRIPTDIR in asgsh
In `asgsh`, when reloading a profile with `rl`, the ASGS Shell checks to see if `$PWD` and `$SCRIPTDIR` are the same, and if not, emits an error message and resets `SCRIPTDIR=$PWD` thusly: 
```
asgs (default-asgs)> pwd
/work/00976/jgflemin/ls6/asgs
asgs (default-asgs)> rl
(info) loaded 'default-asgs' into current profile
asgs (default-asgs)> cd ..
asgs (default-asgs)> rl
(!! warning) SCRIPTDIR is not the same as your PWD, '/work/00976/jgflemin/ls6'
(!! warning)  ... reseting ... maybe 'clone profile' to avoid this warning in the future ...
(info) loaded 'default-asgs' into current profile
```
The code for this is in `cloud/general/DOT-asgs-brew.sh` in the `load` function (near line 251):
```
       # check SCRIPTDIR
        if [ "$SCRIPTDIR" != $(pwd) ]; then
          echo "$W SCRIPTDIR is not the same as your PWD, '$(pwd)'"
          echo "$W  ... reseting ... maybe 'clone profile' to avoid this warning in the future ..."
          export SCRIPTDIR=$(pwd)
        fi
```
Not sure how to address this, I want to fix it but don't want to affect the existing use case.

jasonfleming:
resolves #864

### DIFF
--- a/cloud/general/DOT-asgs-brew.sh
+++ b/cloud/general/DOT-asgs-brew.sh
@@ -248,12 +248,6 @@ load() {
         export _ASGSH_CURRENT_PROFILE="$NAME"
         _reset_ephemeral_envars
        source "$ASGS_META_DIR/$NAME"
-        # check SCRIPTDIR
-        if [ "$SCRIPTDIR" != $(pwd) ]; then
-          echo "$W SCRIPTDIR is not the same as your PWD, '$(pwd)'"
-          echo "$W  ... reseting ... maybe 'clone profile' to avoid this warning in the future ..."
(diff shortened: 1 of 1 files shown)


wwlwpd (approved):
Thank you, I totally forgot about this source of confusion.

PR #900, opened 2022-07-09
Default to tacc_tds3 at TACC and fixed ls6 as cp capable instead of lonestar6

### LINKED ISSUE #899: Make tacc_tds2 the default thredds server at TACC instead of tacc_tds
The platform configurations at TACC are set to post to `tacc_tds` which is the now-defunct `adcircvis.tacc.utexas.edu` virtual machine. These need to be changed to `tacc_tds2` (which is `chg-1.oden.tacc.utexas.edu`) by default, preserving the option for Operators to use `tacc_tds3` (which is also the `chg-1.oden` virtual machine except that is uses `scp` for transfers).

jasonfleming:
resolves #899


wwlwpd:
How do you feel about changing everything to use `tacc_tds3` - since it's the same and I just added that to have something to add. This is the "new" way to add TDS servers. (supports existing way of using it in a config, `TDS=( tacc_tds3 )` )

https://github.com/StormSurgeLive/asgs/blob/master/ssh-servers/tacc_tds3.sh

jasonfleming:
Works for me, I will make this change now.

jasonfleming:
ok I have made this tacc_tds3 instead, please have another look :-)

wwlwpd (approved):
looks good to me, going to merge manually due to need to squash

wwlwpd (approved):
looks good to me

wwlwpd:
squashed and cherry-picked manually, closing but this has been merged.

PR #902, opened 2022-07-09
added _unset_statevars function and inserted calls

### LINKED ISSUE #901: asgsh does not unset state variables when statefile is not found
When starting over with an asgs instance by removing the old statefile, the `asgsh` retains the old values of `RUNDIR`, `STATEFILE`, and `SYSLOG`:
```
asgs (HSOFS_nam_jgf)> show statefile
(info) STATEFILE is defined as '/scratch/00976/jgflemin/HSOFS_nam_jgf.state'
asgs (HSOFS_nam_jgf)> rotateRUNDIR.sh /scratch/00976/jgflemin/HSOFS_nam_jgf.state
asgs (HSOFS_nam_jgf)> rl
(info) loaded 'HSOFS_nam_jgf' into current profile
(info) config file found, instance name is 'HSOFS_nam_jgf'
(!! warning) state file '/scratch/00976/jgflemin/HSOFS_nam_jgf.state' does not exist.
(info) no indication of first run yet?
asgs (HSOFS_nam_jgf)> show statefile
(info) STATEFILE is defined as '/scratch/00976/jgflemin/HSOFS_nam_jgf.state'
asgs (HSOFS_nam_jgf)> show syslog
(info) SYSLOG is defined as '/work/00976/jgflemin/ls6/log/HSOFS_nam_jgf.asgs-2022-Jul-09-T11:55:12-0500.2422845.log'                                                                                                           
asgs (HSOFS_nam_jgf)> show rundir
(info) RUNDIR is defined as '/scratch/00976/jgflemin/asgs2422845'
```
As a result, the restarted ASGS instance will try to write to the old run directory and the old syslog etc.

jasonfleming:
resolves #901

### DIFF
--- a/cloud/general/DOT-asgs-brew.sh
+++ b/cloud/general/DOT-asgs-brew.sh
@@ -285,11 +285,11 @@ _reset_ephemeral_envars() {
 # used to unset variables stored in statefile when
 # the statefile is not found or is moved or deleted
 _unset_statevars() {
-  export STATEFILE=
-  export RUNDIR=
-  export LASTSUBDIR=
-  export SYSLOG=
-  export ADVISORY=
+  unset STATEFILE
+  unset RUNDIR
+  unset LASTSUBDIR
+  unset SYSLOG
+  unset ADVISORY
 }
 
 _parse_config() {
@@ -670,8 +670,7 @@ delete() {
      read -p "This will delete the state file, \"${STATEFILE}\". Type 'y' to proceed. [N] " DELETE_STATEFILE
      if [[ 'y' == "${DELETE_STATEFILE}" ]]; then
        rm -rvf "${STATEFILE}"
-       export STATEFILE=
-       _unset_statevars
+       _unset_statevars # also unsets STATEFILE
      else
        echo "Purge of state file cancelled."
      fi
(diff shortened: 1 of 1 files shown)


cloud/general/DOT-asgs-brew.sh
```diff
@@ -282,6 +282,16 @@ _reset_ephemeral_envars() {
   export ASGS_CONFIG=
 }
 
+# used to unset variables stored in statefile when
+# the statefile is not found or is moved or deleted
+_unset_statevars() {
+  export STATEFILE=
```
wwlwpd, line 288: Is there a reason that `unset` would not be used here? I do like the idea of cleaning things up like this; and I am not exactly sure of the side effect of `unset VAR` versus `export VAR=`. Did you try `unset` and find it was causing issues?

wwlwpd (commented):
I like the idea. I would like to determine if `unset` is more appropriate here.

wwlwpd:
It looks like if we use `unset`, we can also start to get more formal about checking if the variable exists at all versus is set to an empty string,

https://stackoverflow.com/questions/12262696/using-unset-vs-setting-a-variable-to-empty

Sorry I am being so pedantic about this, but getting this right will allow us to provide an even tighter environment incrementally. Thanks!

cloud/general/DOT-asgs-brew.sh
jasonfleming: I used `export VAR=` to keep this consistent with the existing implementation of `_reset_ephemeral_envars()`'  I thought about using `unset` instead but was not sure if the unsetting would propagate to a subshell, and it seemed that `export VAR=` definitely would. However, I just did a little command line test and found that the unsetting does seem to propagate. So I will change this from `export VAR=` to `unset` and test.

jasonfleming:
Hey @wwlwpd I made the changes to `unset`. Please have another look. :-)

wwlwpd (approved):
looks like a great improvement to the enviroment, thank you!

wwlwpd:
Merged manually, with a squash and cherry-pick.

PR #903, opened 2022-07-09
Added support for "postAdditionalFiles" in ASGS configs.

### LINKED ISSUE #894: add ASGS config variable for specifying additional files to post when using `opendap_post2.sh`

wwlwpd:
Issue 894: "postAdditionalFiles" can be defined as a bash array in ASGS
configuration files for defining custom files to be added to the file list
that is transfered by output/opendap_post2.sh.

Example in context of an ASGS configuration file,

  OPENDAPPOST=opendap_post2.sh
  POSTPROCESS=( includeWind10m.sh createOPeNDAPFileList.sh "${OPENDAPPOST}" )
  postAdditionalFiles=(file1.ext file2.ext fileN.etc) #<~ custom files to send via $OPENDAPPOST
  # ^^ new variable ^^

The basic process is that writeProperties() in ./writeProperties.sh writes
to run.properties, and from there output/opendap_post2.sh gets this list and
sends the files.

NOTE: This change includes some refactors that eliminated loops using ${ARRAY[*]}
into single line produced the desired effect, "echo ${ARRAY[@]}".

Resolves #894.


wwlwpd:
can't figure out how to add @notstarboard as a 2nd reviewer; hopefully this mention gives some visibility

output/opendap_post2.sh
```diff
@@ -77,9 +89,29 @@ if [[ ${#SERVERS[@]} -eq 0 ]]; then
...
+IFS=' ' read -r -a FILES_Standard <<< "$fileList"
+
+declare -a FILES_Additional
+additionalFileList=${properties["post.opendap.additionalFiles"]}
+IFS=' ' read -r -a FILES_Additional <<< "$additionalFileList"
+
 declare -a FILES
-fileList=${properties["post.opendap.files"]} # array of files to post to opendap
-IFS=' ' read -r -a FILES <<< "$fileList"  # FIXME: contains "(" and ")" (don't use eval)
+FILES=( "${FILES_Standard[@]}" "${FILES_Additional[@]}" ) # NOTE: duplicates are not removed or detected
```
jasonfleming, line 109: I tested this on the  command line and found that the parentheses were retained due to the reasons described above for `SERVERS`: 
```
asgs (HSOFS_gfs_jgf)> echo ${FILES[@]}                                                                                                                                                                                          
( fort.15 fort.22 fort.61.nc fort.63.nc maxele.63.nc run.properties fort.74.nc maxwvel.63.nc sendNotification endrisinginun.63.nc everdried.63.nc fort.64.nc fort.68.nc fort.71.nc fort.72.nc fort.73.nc initiallydry.63.nc inun
dationtime.63.nc maxinundepth.63.nc maxrs.63.nc maxvel.63.nc minpr.63.nc /scratch/00976/jgflemin/asgs2369694/2022071000/cycle.log /scratch/00976/jgflemin/asgs2369694/2022071000/nowcast/scenario.log /work/00976/jgflemin/ls6/a
sgs-configs/2022/asgs_config_HSOFS_gfs_jgf.sh /work/00976/jgflemin/ls6/log/HSOFS_gfs_jgf.asgs-2022-Jul-09-T11:19:43-0500.2369694.log ) ( file1 file2 fileN )
```

output/opendap_post2.sh
```diff
@@ -77,9 +89,29 @@ if [[ ${#SERVERS[@]} -eq 0 ]]; then
...
+
+#+ vvvvv
+# BEGIN - determine what files to upload, takes into
+# consideration files written in the scenario's run.properties
+# file, using 2 keys:
+#   + post.opendap.files           - generated by output/createOPeNDAPFileList.sh
+#   + post.opendap.additionalFiles - specified by user config via "postAdditionalFiles"
+#
+declare -a FILES_Standard
+fileList=${properties["post.opendap.files"]}
```
jasonfleming, line 101: For reasons described above, I suggest
```
declare -a fileList
fileList=( $( echo ${properties["post.opendap.files"]} | tr -d "()") )
```
Not sure the `declare` is even needed.

output/opendap_post2.sh
```diff
@@ -77,9 +89,29 @@ if [[ ${#SERVERS[@]} -eq 0 ]]; then
...
+IFS=' ' read -r -a FILES_Standard <<< "$fileList"
+
+declare -a FILES_Additional
```
jasonfleming, line 104: For reasons described above, I suggest
```
declare -a additionalFileList
additionalFileList=( $( echo ${properties["post.opendap.files"]} | tr -d "()") )
```
Not sure the declare is even needed.

output/opendap_post2.sh
```diff
@@ -77,9 +89,29 @@ if [[ ${#SERVERS[@]} -eq 0 ]]; then
...
+additionalFileList=${properties["post.opendap.additionalFiles"]}
```
jasonfleming, line 105: The `post.opendap.files` property only consists of files that are confirmed to exist (at least), because the script that creates the list checks with either `ls` or `-e` (in all cases?). But this file list has been formed without confirming that the files in the list both exist and are not zero length. Not sure what is the best way to handle this. Maybe add a check and emit a warning message if they are not found or are zero length? And then remove them from the list of files to be transferred?

jasonfleming (commented):
Hey @wwlwpd I tested this on the command line and have some suggested alt formulations for extracting and formulating the lists of files to post.

wwlwpd:
Thanks for the comments, will update according to your suggestions. I will look at checking that the additional files also exist (do you think the zero size is needed? I'd imagine it'd be good to send them anyway? Idk - thoughts?)

output/opendap_post2.sh
wwlwpd: FWIW, you (assuming) have accounted for this where `$FILES` are looped over,

```
      while [[ $fileIndex -lt `expr ${#FILES[@]} - 1` ]] ; do  # skip the closing "("
         file="${FILES[$fileIndex]}"
         if [[ $file = '(' || $file = ')' || ! -s $file ]]; then
            fileIndex=`expr $fileIndex + 1` 2>> $SCENARIOLOG
            continue
         fi
 ...
```

wwlwpd:
@jasonfleming - all changes since last review were made in `output/opendap_post2.sh` - cleaned up a bunch of stuff as a result. I think the only thing I didn't do was check for size 0; but only files that exist are uploaded. In additional to this, all files are resolved to their full paths upfront. I also added a `-w` check in the `SSHCMD` that has the `chmod` - checks to see if that file is writable by `$USER`.

jasonfleming (approved):
This looks great @wwlwpd thank you.

PR #912, opened 2022-07-10
Allowed grib2 files to be reported as type 'data' from bash file command

### LINKED ISSUE #911: bash file command v5.11 on supermic reports grib2 file type as 'data'
The `bash` `file` command is used to smoke test locally downloaded grib2 files. Newer versions of `file` report the file type of grib2 files as `(GRIB) version 2` whereas older versions (e.g., v5.11 on supermic) report the type of this file as simply `data`.

jasonfleming:
resolves #911

### DIFF
--- a/downloadGFS.sh
+++ b/downloadGFS.sh
@@ -164,12 +164,12 @@ downloadGFS()
             # (may want to turn off this conditional for end-to-end testing)
             subsetSuccess=0
             localFile=$instanceGfsDir/$yyyymmdd/$file
+            cmd="curl -s "$finalURL" > $localFile 2>> $SYSLOG" # used in error messages
             while [[ $subsetSuccess -eq 0 ]]; do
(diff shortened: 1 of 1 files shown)


downloadGFS.sh
```diff
@@ -183,7 +183,7 @@ downloadGFS()
                     localFileType="$(file $localFile 2>> $SYSLOG)"
```
wwlwpd, line 183: I am going to approve and merge this, but maybe we need a separate utility that does some deep verification if this step is actually needed. E.g., if we're just checking that it's "data" then that doesn't seem to be much different than making sure the file exists or that it's got a size > 0 bytes. Just a thought!

wwlwpd (approved):
minor change, adds an additional check but maybe we need some deeper checks in the future - probably best to just wait it out and see.

PR #918, opened 2022-07-12
Updated RMQ message to name correct wind product, confirmed asgs-msgr…

notstarboard:
This should address issue 884, where asgs-msgr.sh and asgs-msgr.py both use /usr/bin/python, which is the system python 2 (at least on hatteras). Now it will use python3, which should always be the ASGS python 3 when run via asgsh.

I also snuck in a change to an RMQ message in downloadGFS.sh that was confusing me for a bit in the console output. Now it should rightly print "GFS" when downloading GFS data instead of "NAM".

---

I've run into no errors from any of these changes, but I should caveat this with the fact that I'm still struggling to get my ASGS instance working as expected. If these changes introduced problems, depending on how subtle, it could be harder for me to catch them at present.

### DIFF
--- a/monitoring/asgs-msgr.sh
+++ b/monitoring/asgs-msgr.sh
@@ -4,7 +4,7 @@ THIS=$(basename -- $0)
 
 echo "RMQMessaging:Validating Message Service..."
 
-RMQMessaging_Python=$(which python3); # should be using python provided by ASGS environment
+RMQMessaging_Python=python3 ; # should be using python provided by ASGS environment
 
 if [[ ! -e ${RMQMessaging_Python} ]] ; then 
 	echo "Specified Python (${RMQMessaging_Python}) not found. Turning off RMQMessaging."
--- a/monitoring/asgs-msgr.sh
+++ b/monitoring/asgs-msgr.sh
@@ -4,7 +4,7 @@ THIS=$(basename -- $0)
 
 echo "RMQMessaging:Validating Message Service..."
 
-RMQMessaging_Python=$(which python); # should be using python provided by ASGS environment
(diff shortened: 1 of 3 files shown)


monitoring/asgs-msgr.sh
```diff
@@ -4,7 +4,7 @@ THIS=$(basename -- $0)
 
 echo "RMQMessaging:Validating Message Service..."
 
-RMQMessaging_Python=$(which python); # should be using python provided by ASGS environment
+RMQMessaging_Python=$(which python3); # should be using python provided by ASGS environment
```
wwlwpd, line 7: this should just be,

```
RMQMessaging_Python=python3
```

Because this is already available via `asgsh`'s `PATH` environmental variable.

wwlwpd (changes requested):
Thanks for the PR!! I'd like to treat this PR as one that "gets RMQ" working in ASGS with `python3` - which is different than how we normally treat things. However since it's currently broken I don't want there to be incremental changes that don't get it fully working again. I can help with changing the bits that install the required modules during `./init-asgs.sh` (or after); but assuming the modules are installed this just needs to be able to work as before. Let me know clarification is needed. Thanks, again!

wwlwpd:
I moved this to draft so that it's easier to check the status in the PR list. When ready pls mark it "Ready for Review". Thanks!

notstarboard:
> I moved this to draft so that it's easier to check the status in the PR list. When ready pls mark it "Ready for Review". Thanks!

Apologies, I made the change but I didn't know to mark it ready for review! Should be all set.

wwlwpd (approved):
Looks good to me. I will take care of squashing these 3 commits into 1.

PR #921, opened 2022-07-14
Adding "sendNotification" file tracer back opendap_post2.sh.

### LINKED ISSUE #919: emails broken again
Probably related to a mistake I made in #894.

wwlwpd:
Issue 919: During the refactor of opendap_post2.sh, I missed
the email comms tracer, "sendNotification" - which is meant to
act as signal for the script to send an email to the addresses
specified in the configuration file. This was getting removed bec
it isn't meant to be an actual file on the filesystem.

If detected "sendNotification" is added as the last item in the
list of files, so that the message is sent out after all of the
files have been processed. This may provide an opportunity in
the future to have more robust checks and messaging.

Resolves #919.

### DIFF
--- a/output/opendap_post2.sh
+++ b/output/opendap_post2.sh
@@ -109,7 +109,14 @@ _FILES=( "${FILES_Standard[@]}" "${FILES_Additional[@]}" ) # NOTE: duplicates ar
 
 # now filter out any files that do not exist
 FILES=()
+sendNotification=
 for _file in ${_FILES[*]}; do
+  # check for 'sendNotification' tracer before `readlink -f` which requires a real file
+  if [[ $_file == "sendNotification" ]]; then
+    sendNotification=$_file
+    FILES+=($_file)
+    continue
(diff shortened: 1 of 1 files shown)


jasonfleming (changes requested):
Hey @wwlwpd it is good to detect this `sendNotification` faux file name and handle it even though it does not exist. But the purpose of it is to send the notification email early (before all the files have been uploaded) to avoid delaying the sending of the notification email for for files that will not be needed by the recipient(s) (e.g., CERA). The way the script currently works, If `sendNotification` is not in the list of files, the notification is automatically sent after all files have been transmitted. I think the `if` statement on line 115 could be changed to `if [[ -e $_file || $_file == "sendNotification" ]]; then`

wwlwpd:
I will take care of that ordering issue with the tracer.

jasonfleming (approved):
This looks great, thank you @wwlwpd

PR #932, opened 2022-07-21
COAMPS-TC support

### LINKED ISSUE #850: Add support for COAMPS-TC

jasonfleming:
resolves #850

### DIFF
--- a/downloadCOAMPSTC.sh
+++ b/downloadCOAMPSTC.sh
@@ -1,0 +1,490 @@
+#!/bin/bash
+
+#trap read debug
+#----------------------------------------------------------------
+# downloadCOAMPSTC.sh: Subroutine that polls the MetGet service
+# to determine status of COAMPS-TC data and then subsets
+# and downloads it via web client.
+#----------------------------------------------------------------
+# Copyright(C) 2022 Jason Fleming
+#
+# This file is part of the ADCIRC Surge Guidance System (ASGS).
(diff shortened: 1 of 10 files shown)


downloadCOAMPSTC.sh
```diff
@@ -0,0 +1,490 @@
+#!/bin/bash
```
wwlwpd, line 1: We need to start putting all scripts in `./bin` or somewhere else that's in `PATH`. If it's not in `PATH`, then we need to add it there. `$SCRIPTDIR` is not currently in `PATH` but probably should be just for the existing scripts there.

wwlwpd (changes requested):
Big request atm: pls put `downloadCOAMPSTC.sh` into `./bin` or some other directory that's in `PATH`. If you need to create a new `bin` somewhere and it should be in `PATH`, this is controlled in `./cloud/general/asgs-brew.pl`.

downloadCOAMPSTC.sh
jasonfleming: can do

jasonfleming:
Hey @wwlwpd before I move `downloadCOAMPSTC.sh` to `/bin`, I wanted to point out that this script is not meant to be executed as a standalone, it is meant to be `source`d into `asgs_main.sh` ... I didn't know if this changes your mind about it

PR #942, opened 2022-07-22
critical path fix for copying Wind10m layers rather than using ln -s

### LINKED ISSUE #940: `includeWind10m.sh` uses symbolic links, which confuses `opendap_post2.sh` `readlink` path resolver

jasonfleming:
resolves #940 ; testing this now


wwlwpd (approved):
Good catch, thanks for the fix!

Issue #954, opened 2022-07-26
RMQMessaging Still Not Working Post-Issue-918 Fix


notstarboard:
My Issue-918 fix didn't get RMQ messaging running - I should have reformulated the if check now that we're not using which. As is, the check fails to recognize when python3 exists. I'll get going on the necessary change now.

PR #956, opened 2022-07-26
Fixed python3 existence check

notstarboard:
This is to address Issue #954


monitoring/asgs-msgr.sh
```diff
@@ -6,7 +6,7 @@ echo "RMQMessaging:Validating Message Service..."
 
 RMQMessaging_Python=python3 ; # should be using python provided by ASGS environment
 
-if [[ ! -e ${RMQMessaging_Python} ]] ; then 
+if [[ ! -e $(command -v ${RMQMessaging_Python}) ]] ; then
```
wwlwpd, line 9: I'd like use `which` here, mainly for consistency. But I also had to look up what `command -v` does, and it seems like `which` is what we want. 

But that does give me an idea for a `better` command that acts like `which` so I can say, "`which` is `better`" without it being a question. xD

... on second thought, change it also to use `-x` and wrap it in quotes, though it'd be strange for a platform to not have `python3` these days:

```
 if [[ ! -x "$(which ${RMQMessaging_Python})" ]] ; then 
```

This is untested, please make sure it works before you update the PR. Thanks!

wwlwpd (changes requested):
see comment for requested change

notstarboard:
which and -x seem to work just fine!

wwlwpd (approved):
Looks good to me!

wwlwpd:
I approved it  but will merge it later since it has to be squashed.

wwlwpd:
merged this with a `cherry-pick` externally due to _squash_

PR #959, opened 2022-07-28
Adding support for local, custom ADCIRC src.

### LINKED ISSUE #748: support adding untracked ADCIRC patchsets
Also generate menu options dynamically based on contents of the directories contained in `$ADCIRC_PATCH_PATH`.

wwlwpd:
Issue-748: This will allow users with ADCIRC source somewhere on their
ASGS host machine to build and register ADCIRC via `build adcirc`, by
selecting the "custom" option and answering the questions in the interactive
mode that follows. Upon successful compilation, this ADCIRC is available
like all others via "load adcirc" and can be seen along with the others
via "list adcirc". Patches are not supported at this time.

Resolves #748.

### DIFF
--- a/bin/init-adcirc.sh
+++ b/bin/init-adcirc.sh
@@ -25,14 +25,16 @@
 # before running 'build adcirc' in asgsh
 #
 
+I="(info)"
+W="(!! warning)"
+
 if [ "${1}" = "clean" ]; then
   echo "'clean' not implemented for optional ADCIRC/SWAN step at this time, clean up for ADCIRC and SWAN must be done manually."
   exit 0
 fi
 
-
 ADCIRCS=()
-NUM_ADC=${#ADCIRCS[@]}
+NUM_ADC=0
 
 # support function for menu detecting selection
 _is_a_num()
@@ -55,20 +57,24 @@ _show_supported_versions()
   echo  '|~~~VERSION~~~~~~~~~~~~~~~~~~~~~~~~~~~~~DESCRIPTION'
   for VERSION in $(ls -1 $SCRIPTDIR/patches/ADCIRC); do 
     _ADCIRCS="$_ADCIRCS $VERSION"
-    local about="patchset for Version,$VERSION"
+    local about="patchset for Version, $VERSION"
     if [ -e $SCRIPTDIR/patches/ADCIRC/$VERSION/about.txt ]; then
       about=$(cat $SCRIPTDIR/patches/ADCIRC/$VERSION/about.txt | sed 's/\n//g')
     fi
     num=$(($num+1))
     printf "|%-2s) %-33s | %-66s |\n" $num $VERSION "$about"
   done
+    # final menu entry for custom directory
+    num=$(($num+1))
+    printf "|%-2s) %-33s | %-66s |\n" $num custom "select this option for custom directory"
   echo  "--"
   echo
-  ADCIRCS=($_ADCIRCS)
+  ADCIRCS=($_ADCIRCS custom)
   if [ "${1}" != "noexit" ]; then
     # exits on error if '1' is optionally passed, defaults to 0 (no error)
     exit ${1:-0} 
   fi
+  NUM_ADC=$num
 }
 
 if [ "${1}" = "supported" ]; then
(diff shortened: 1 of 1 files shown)


wwlwpd:
@jasonfleming this needs to be tested _and_reviewed. @notstarboard should also take a look at it. Let me know when y'all can discuss and I can describe the testing procedure; really just requires local file access to the original "source" (the original source is left untouched, just `cp`'d).

jasonfleming:
Will work on this on hatteras; need a copy of the source code ; copy to work dir where asgs is setup ; start shell ; run build adcirc ; last menu option is custom ; (11 for now) ; will ask for path to source ; tld  not work ; ask for name etc ; skips patching and git checkout ; copies source from you dir to normal asgs place ; from there identical to previous procedure ; adds to adcirc meta dir file with info needed ; also produces json file ; new key called `custom_src` plus env var ; if it is pointing to custom source this has full path ; otherwise 0 value ; once built successfully works just like any other adcirc build ; `load adcirc x` etc ; will get path to source via email

wwlwpd:
rebased on current master SHA, f3f89d82687802d88f79a301213abc73c19aa9b2

jasonfleming:
Thank you @wwlwpd I will be testing this today and providing feedback.

wwlwpd:
@dave-gso has reported an interest issue, I will try to reproduce and fix if found.

bin/init-adcirc.sh
```diff
@@ -190,38 +239,56 @@ ADCSWAN_MAKE_CMD="make $ADCSWAN_BINS compiler=${ADCIRC_COMPILER} MACHINENAME=${A
...
+# git repo wrangling if not copying from a local filesystem
+#TODO: put all this in function
+if [[ -d "${CUSTOM_SRC}" ]]; then                     # handle custom filesystem based source, no patches
+  # copy across local file system
+  mkdir -p $ADCIRCBASE
+  echo " ... copying from from '$CUSTOM_SRC' to '$ADCIRCBASE'"
+  cp -rf $CUSTOM_SRC/* $ADCIRCBASE
+  # clean up destination
+  echo " ... cleaning up cruft from '$ADCIRCBASE'"
+  rm -rf $ADCIRCBASE/.git                2> /dev/null
```
jasonfleming, line 251: suggest to use `make clean` or similar

jasonfleming (dismissed):
This looks great @wwlwpd thank you. Suggest additional comments to document any assumptions or caveats, e.g., that the `make` build method is used (not cmake), and that the netcdf paths should not be hard coded in the developers' build system.

bin/init-adcirc.sh
wwlwpd: good idea, will do!

wwlwpd:
@jasonfleming I will commit to completing #975, to document `build adcirc` in the wiki once this is merged - just to keep the scope of the work down.

wwlwpd:
Latest commit,

```
commit 708c8328a3bff0573e5c125b0359246f1264074c
Author: wwlwpd <46434714+wwlwpd@users.noreply.github.com>
Date:   Fri Aug 12 04:09:28 2022 -0400

    fixup! Adding support for custom paths to ADCIRC src.
    
    Replaced series of `rm -rv` with `make clobber`; detects if
    target directory exists, and if so will offer to delete it. The
    default for this is set to N (no). Added fun Star Wars ASCII art
    that is shown whenever ADCIRC build succeeds.
```

wwlwpd:
Preliminary documentatio, https://github.com/StormSurgeLive/asgs/wiki/ASGS-Compiling-ADCIRC

jasonfleming (approved):
Hey @wwlwpd I tested this on hatteras, and was able to build @dave-gso weir_pump code successfully. I executed `adcirc` on the command line in the asgs shell and it was able to start successfully without any library errors. I was then able to build v55.01 (bonus testing) successfully and test run `adcirc` on the command line. Looks great!

wwlwpd:
Thanks for testing, will squash and merge!

wwlwpd:
Merged manually. Approved by @jasonfleming

PR #969, opened 2022-08-04
Added way to prepend a directory to the OpenDAP path formation.

### LINKED ISSUE #931: `opendap_post2.sh` needs some way to specify that the base branch on the remote server is not the same as an operational run
currently "replay" runs are getting shoved into the same `$YEAR` as any other operational run

wwlwpd:
Issue 931: Defining OPENDAPADDROOT in an ASGS configuration file will
cause an additional directory to be prepended to the data path
created and used as the upload target, in output/opendap_post2.sh.

For example, adding the following to the ASGS config fle,

```
  OPENDAPADDROOT=replay-test
```

Will ultimately make the files accessible in a url that is rooted
at 'replay-test', e.g.,

```
  http://chg-1.oden.tacc.utexas.edu/thredds/catalog/asgs/replay-test/2022/al05/11/TX2008/ls6.tacc.utexas.edu/TX2008-replay-ls6/nhcConsensus/catalog.html
                                                         ^^^^^^^^^^^^
```

If not set, the URL would normally look as follows,

```
  http://chg-1.oden.tacc.utexas.edu/thredds/catalog/asgs/2022/al05/11/TX2008/ls6.tacc.utexas.edu/TX2008-replay-ls6/nhcConsensus/catalog.html
                                                         ^^^^^^^^^^^^
```
Resolves #931.

### DIFF
--- a/output/opendap_post2.sh
+++ b/output/opendap_post2.sh
@@ -225,6 +225,7 @@ for server in ${SERVERS[*]}; do
    DOWNLOADPREFIX=${properties["post.opendap.${server}.downloadprefix"]}
    CATALOGPREFIX=${properties["post.opendap.${server}.catalogprefix"]}
    OPENDAPBASEDIR=${properties["post.opendap.${server}.opendapbasedir"]}
+   OPENDAPADDROOT=${properties["post.opendap.addroot"]}
    #
    #--------------------------------------------------------------------
    #  O P E N  D A P    P A T H   F O R M A T I O N
@@ -235,7 +236,11 @@ for server in ${SERVERS[*]}; do
       # for NAM, the "advisory number" is actually the cycle time
       YEAR=${properties["forcing.nwp.year"]}
       NWPMODEL=${properties["forcing.nwp.model"]}
-      STORMNAMEPATH=$YEAR/$NWPMODEL
+      if [ -n "${OPENDAPADDROOT}" ]; then
+        STORMNAMEPATH=$OPENDAPADDROOT/$YEAR/$NWPMODEL
+      else
+        STORMNAMEPATH=$YEAR/$NWPMODEL
+      fi
    fi
    if [[ $TROPICALCYCLONE = on ]]; then
       YEAR=${properties["forcing.tropicalcyclone.year"]}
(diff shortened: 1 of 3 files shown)


wwlwpd:
Looking for some early feedback, also in the process of testing this.

jasonfleming (commented):
This is great @wwlwpd we definitely need this to disambiguate replay storms vs actual on the destination server. Probably has a lot of applications beyond that as well. I know this is still a draft, but looking at the code changes in `opendap_post2.sh`, it looks like `$STORMNAMEPATH` will be set to e.g. `/$YEAR/$NWPMODEL` instead of `$YEAR/$NWPMODEL` if `$OPENDAPROOT` is not set. Not sure how much of an issue this is.

wwlwpd:
> This is great @wwlwpd we definitely need this to disambiguate replay storms vs actual on the destination server. Probably has a lot of applications beyond that as well. I know this is still a draft, but looking at the code changes in `opendap_post2.sh`, it looks like `$STORMNAMEPATH` will be set to e.g. `/$YEAR/$NWPMODEL` instead of `$YEAR/$NWPMODEL` if `$OPENDAPROOT` is not set. Not sure how much of an issue this is.

I was just going for inserting a directory in the root as opposed to dealing with any logical view of the directory structure. I found an error in running and have fixed it, though. I ended up writing it to `run.properties` and pulling it out directly using the associative array in the `opendap_post2.sh` that is generated from reading that file on start. Switching from draft to ready for review.

jasonfleming:
Hey @wwlwpd if `$OPENDAPADDROOT` is not set, then will the `$STORMNAMEPATH` end up being something like `null/$YEAR/$NWPMODEL`?

wwlwpd:
This is the error I was seeing originally. I will make sure this won't happen if not set.

wwlwpd:
> Hey @wwlwpd if `$OPENDAPADDROOT` is not set, then will the `$STORMNAMEPATH` end up being something like `null/$YEAR/$NWPMODEL`?

After considering this question more, if it is set literally to `null` it'll currently use that string. In order to be ignored it must be set to nothing; either not exist in the ASGS config or be set like,

```
OPENDAPADDROOT=
```
Then it won't get written to `run.properties` and therefore won't get run. I'd prefer to not treat `null` as a special value. Don't read this to mean that I want to refactor everything, but I think it was a mistake to treat _null_ as a special value and I don't want to add it here, if that's okay with you. Again, this doesn't me to change this anywhere it exists. It just means I want to take the step of stopping to do it. :)

jasonfleming:
Thanks @wwlwpd, sounds good. I originally started setting shell variables to `null` because long command line argument lists got messed up if any of the variables was unset ... and it was very hard to debug it when it happened. So I decided to always make sure that variables were set to something. Not wedded to this approach at all, and I welcome new ideas on this.

jasonfleming:
Should I approve this now or wait until you have it in final form?

wwlwpd:
If it looks okay, feel free to approve it. I will not merge it until it cycles through a few more replay advisories on my test run (tx2008 @ tacc).

output/opendap_post2.sh
```diff
@@ -225,35 +225,41 @@ for server in ${SERVERS[*]}; do
    DOWNLOADPREFIX=${properties["post.opendap.${server}.downloadprefix"]}
    CATALOGPREFIX=${properties["post.opendap.${server}.catalogprefix"]}
    OPENDAPBASEDIR=${properties["post.opendap.${server}.opendapbasedir"]}
+   OPENDAPADDROOT=${properties["post.opendap.addroot"]}
    #
    #--------------------------------------------------------------------
    #  O P E N  D A P    P A T H   F O R M A T I O N
    #--------------------------------------------------------------------
    STORMNAMEPATH=null
```
jasonfleming, line 233: Should this be `STORMNAMEPATH=`?

wwlwpd: Good question. I didn't change it so should still be functioning as it was originally. I will take a closer look.

wwlwpd:
> Thanks @wwlwpd, sounds good. I originally started setting shell variables to `null` because long command line argument lists got messed up if any of the variables was unset ... and it was very hard to debug it when it happened. So I decided to always make sure that variables were set to something. Not wedded to this approach at all, and I welcome new ideas on this.

Good info, exactly the reason why I am not interested in just changing everything. I assume everything has been done with some careful consideration :).

output/opendap_post2.sh
wwlwpd: > Should this be `STORMNAMEPATH=`?

Do you mean should we just expose `STORMNAMEPATH`'s initial value in the configuration? I re-read this as that and it might not be a bad idea. Basically rather than have `OPENDAPADDPATH`, just have it be `STORMNAMEPATH` - or do we want to combine then into a single exposed variable that sounds more like what it is, e.g., `OPENDAPROOTPATH`? (which is what I had been playing with originally).

wwlwpd:
@jasonfleming I redid things and tested under the following circumstances:

* `OPENDAPADDROOT` is not set  - `post.opendapp.addroot` won't exist in `run.properties`
* `OPENDAPADDROOT=some-string` - `post.opendap.addroot` is set to `some-string` in `run.properties`
* `OPENDAPADDROOT=` - `post.opendap.addroot` won't exist in `run.properties`

In all cases, it does what is expected.

I am running a final test now, but I am hopefull this works as intended.

wwlwpd:
This looks to be working for me, once approved I will merge this and also update the Wiki (extended documentation on postprocessing)

wwlwpd:
@jasonfleming @notstarboard

jasonfleming (approved):
This looks great, thank you @wwlwpd! Could add an initialization with `OPENDAPADDROOT=` in `variables_init.sh`. :-)

wwlwpd:
> This looks great, thank you @wwlwpd! Could add an initialization with `OPENDAPADDROOT=` in `variables_init.sh`. :-)

Done in current version of branch. Merging.

PR #977, opened 2022-08-15
The NHC monitor/recorder was dying whenever the storms.json fetch tim…

### LINKED ISSUE #976: added `alarm` signal handling to `bin/nhc-recorder`
the monitor that grabs advisories was dying unexpectedly whenever the NHC URL to fetch would timeout

wwlwpd:
…ed out.

Issue 976: Weather::NHC::TropicalCyclone's fetch method uses an `alarm`
that is triggered whenever there is a time out. Now defining $SIG{ALRM}
in a "retry" loop.

Added Try::ALRM to list of Perl modules installed.

Resolves #976.

### DIFF
--- a/PERL-MODULES
+++ b/PERL-MODULES
@@ -14,6 +14,7 @@ Params::Validate
 Test::More
 Time::Local
 Try::Tiny
+Try::ALRM
 Template
 Perl::Tidy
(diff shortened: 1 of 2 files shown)


wwlwpd:
After second thought I do not think this is going to be the right fix.

PR #988, opened 2022-08-30
correct swan fort.26 comment lines that were too long

### LINKED ISSUE #987: fix failing SWAN coupling for PRVI on Frontera due to long comment line
Issue #1 has cropped up again for the PRVI mesh on Frontera, and most likely others as well. Looking at this again, it is not clear whether the single quotes are part of the 72 character limit. As a result, I will reduce the string length from 72 to 68 to be certain that we are well within the limit.

jasonfleming:
resolves #987


wwlwpd (approved):
looks good to me!

PR #998, opened 2022-09-09
removed requirement for `fort.88` when `VARFLUX != off`

### LINKED ISSUE #8: initial water level version difference
The ASGS always looks for a fort.88 file if river flux has been turned on. However, this is only appropriate for ADCIRC versions v50 and v49. ADCIRC v51 implements this as a nodal attribute. ASGS should detect the version of ADCIRC and activate the appropriate behavior.

jasonfleming:
resolves #8

The `fort.88` file was used to make sure the nodes on a river influx boundary are wet (otherwise there will be no flow there). This file was supported by ADCIRC v49 and v50, but in v51+ this same capability is provided via nodal attribute. As a result, the requirement for (and support for) a `fort.88` file in ASGS was removed.

### DIFF
--- a/asgs_main.sh
+++ b/asgs_main.sh
@@ -304,14 +304,12 @@ prep()
         fi
     fi
     if [ $START = coldstart ]; then
-       # if we have variable river flux, link the fort.20 and fort.88 files
+       # if we have variable river flux, link the fort.20 file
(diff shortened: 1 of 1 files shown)


wwlwpd (approved):
Looks good to me.

PR #1001, opened 2022-09-14
Added patches to aswip to improve handling troublesome BEST track files

### LINKED ISSUE #997: improve diagnostics, error messages, and general behavior of `aswip`
Sometimes when running the `aswip` (asymmetric wind input preprocessor, bundled with the adcirc source code) we observe an error message like the following:
```
ERROR: aswip:            6  isotachs were nonzero.
```
The root cause of this error message is not known, nor is the practical effect of this error, or what the fix/workaround should be. These should be determined so that we can handle this situation appropriately.

ongoing updates as part of this case: https://github.com/StormSurgeLive/asgs/wiki/Adding-and-Modifying-ADCIRC-Patch-Sets

jasonfleming:
resolves #997

### DIFF
--- a/patches/ADCIRC/v53release-gfortran-desktop/11-aswip-dos2unix.patch
+++ b/patches/ADCIRC/v53release-gfortran-desktop/11-aswip-dos2unix.patch
@@ -1,0 +1,3501 @@
+--- a/wind/aswip_1.0.3.F	2022-09-14 08:38:25.885668737 -0400
++++ b/wind/aswip_1.0.3.F	2022-09-14 08:41:08.569707715 -0400
+@@ -17,7 +17,7 @@
+          USE vortex
+          USE wind, only : writeFullCircleRmaxes, vortexLon,
+      &                 vortexLat, writeRadialVandP,
+-     &                 writeSpatialVandP, nws19get, windDrag, nws20get,
++     &                 writeSpatialVandP, nws19get, windDrag, nws20get,
+      &                 writeSpatialuvp, writeNodaluvp,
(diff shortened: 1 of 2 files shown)


wwlwpd (commented):
Generally speaking, you are taking the correct approach here for adding patches.

wwlwpd:
Any plans to break this out into a full PR ready for review?

jasonfleming:
Hey Brett I am working on this now (for the documentation), but I need advice on version control for aswip (or at least the patches) since it is normally maintained in the ADCIRC repo ... do we establish a different repo, or a fork, or a branch of the ADCIRC repo? If we make patches, do we modify the ADCIRC v53 to make our own aswip then make patches back to all the other versions etc? Seems like a mess, need advice on the cleanest way to keep track.

jasonfleming:
Also I think we need some tests ... want to confirm that improvements to `aswip` (or other scripts that handle ATCF data) are not causing new errors. Need advice on maintainable testing setup.

wwlwpd:
@jasonfleming let's talk about this soon - if you're good with the changes, I can take it off your hands and add it to the right patchset(s).

jasonfleming:
Thank you @wwlwpd I've now added this to the patch management system. One potential issue is that there were line ending issues (as if the `aswip.F` was edited in a Windows-based text editor at some point in the past). So the patch ended up being the whole file in order to fix the line endings. Not sure what to do about that (if anything).

wwlwpd:
Please push the changes to an ADCIRC branch, and I will generate the patches.

jasonfleming:
OK I have made a new branch for this as suggested.

PR #1008, opened 2022-09-29
Local mesh support

wwlwpd:
In principle this works, assuming `$ASGS_LOCAL_DIR` is defined; this is currently accomplished by:

1. having some local asset directory available with `path/to/local-asgs-assets/configs/mesh_defaults` defining meshes
2. mesh assets available per definitions in `path/to/local-asgs-assets/configs/mesh_defaults`
3. this change, assuming `$ASGS_LOCAL_DIR` is configured properly works in `asgsh`'s `list meshes`

Set up for `$ASGS_LOCAL_DIR`, currently requires re-running `./init-asgs.sh` with the `-p path/to/local-asgs-assets` set even if the platform is officially supported. This is because `./update-asgs` and `./asgsh` are where `$ASGS_LOCAL_DIR` are persisted. We need a consistent way to register a local assets directory.

In order to test, we need a test `path/to/local-asgs-assets` directory with a proper `./configs/mesh_defaults.sh` and mesh assets in `./input/meshes/localMeshName/`.

### DIFF
--- a/bin/list
+++ b/bin/list
@@ -40,6 +40,14 @@ _list() {
         printf "% 2d. %s\n" $LISTNUM $m
         LISTNUM=$(($LISTNUM+1))
       done
+      # list locally defined meshes if they exist
+      local LOCAL_MESH_DEFAULTS="${ASGS_LOCAL_DIR}/config/mesh_defaults.sh"
+      if [[ -n "$ASGS_LOCAL_DIR" && -e "$LOCAL_MESH_DEFAULTS" ]]; then
+        for m in $(cat $LOCAL_MESH_DEFAULTS | grep '")' | sed 's/[")]//g' | awk '{print $1}'); do
+          printf "% 2d. %s (** locally defined)\n" $LISTNUM $m
+          LISTNUM=$(($LISTNUM+1))
+        done
+      fi
       ;;
     platforms)
       cat $ASGS_PLATFORMS | egrep '^init_' | sed 's/init_//g' | sed 's/()//g' | awk '{print "- " $1}'
--- a/config/mesh_defaults.sh
+++ b/config/mesh_defaults.sh
@@ -21,7 +21,9 @@
 # along with the ASGS.  If not, see <http://www.gnu.org/licenses/>.
 #----------------------------------------------------------------
 THIS=$(basename -- $0)
-consoleMessage "$THIS: Setting default values for the mesh '${MESH}'."
+source $SCRIPTDIR/monitoring/logging.sh
+consoleMessage "$THIS: Setting default values for the mesh '${GRIDNAME}'."
+
 MESHURL=https://asgs-static-assets.sfo2.digitaloceanspaces.com/meshes
 NODALATTRIBUTESURL=https://asgs-static-assets.sfo2.digitaloceanspaces.com/nodal-attributes
 OFFSETURL=https://asgs-static-assets.sfo2.digitaloceanspaces.com/offsets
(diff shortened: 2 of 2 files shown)


wwlwpd:
Looking for a local use case 😉

wwlwpd:
After testing, before merge, I will need to add documentation on the wiki about setting up a local asset directory

enikidis:
hey! I relly appreciate the help @wwlwpd . Asgs picked up `local-asgs-assets` directory. The next step prompts me to specify the brach I wanted I picked issue-1007. what does that mean for me. What's different with this brach and the "vanilla"  asgs ?

wwlwpd:
> hey! I relly appreciate the help @wwlwpd . Asgs picked up `local-asgs-assets` directory. The next step prompts me to specify the brach I wanted I picked issue-1007. what does that mean for me. What's different with this brach and the "vanilla" asgs ?

@enikidis do you still have the same questions?

wwlwpd:
Here's the documention I added for this PR, https://github.com/StormSurgeLive/asgs/wiki/Adding-Local-Assets

jasonfleming (commented):
These changes are great @wwlwpd @enikidis ... one issue is that I see commits in here that don't seem to be related to local assets (e.g., including the adcirc build info json) ... also, looking at the wiki documentation (which is very much appreciated), it's not clear that the contents of the local assets directory should be copied into the asgs installation directory, since there is an environment variable that points to the local assets directory, and this directory is presumably under separate version control as well as shared among multiple asgs instances ... seems that it would be better to leave it external to the local asgs installation and just point to it with the asgs environment variable ... on that topic, what if the environment variable points to a directory that doesn't exist for whatever reason (whether it is inside the asgs installation dir or elsewhere), need to be able to detect that and emit helpful message.

wwlwpd:
Thanks for the feedback, @jasonfleming. I'll take a look at the commit you're talking about and will clarify the documentation. If I can emit a useful warning message if `ASGS_LOCAL_DIR`.

enikidis:
hey everyone I think I succesfully installed and managed to begin a "run" as I said in todays meeting. I installed this version of asgs and experimental adcirc in seperate directory called "experimental" in my workdir, thinking that the outpout will be posted on this new directory it will be also very helpfull to be able to control the directory where logs,statefiles and the asgsXXXX dirs are. This way outputs will be always seperate from the outputs of the "vanilla" asgs that may "live" one the same machine. On the previous big update of asgs you increased the portability of it a lot and this seems like a logical step.. :)

wwlwpd:
> 

I updated the documentation to assume the local directory is already on the local file system - made the instructions much more clear, also.

wwlwpd:
> on that topic, what if the environment variable points to a directory that doesn't exist for whatever reason (whether it is inside the asgs installation dir or elsewhere), need to be able to detect that and emit helpful message

The best approach would likely be to add a validation step in `init-asgs.sh` and a pre-flight check in `asgsh`. I can update the PR with these things, what do you think?

wwlwpd:
> hey! I relly appreciate the help @wwlwpd . Asgs picked up `local-asgs-assets` directory. The next step prompts me to specify the brach I wanted I picked issue-1007. what does that mean for me. What's different with this brach and the "vanilla" asgs ?

If you don't select a branch, it'll just use what you're on. You're right to use `issue-1007` but if you're already on `issue-1007`, then you don't need to specify anything.. I added that mainly to make it easier for testing, in most cases people can just hit "enter"..

Issue #1004, opened 2022-09-19
support building ADCIRC from local or remote zip file


jasonfleming:

wwlwpd:
I think this just require awareness of `.zip` file and the ability to submit a patch, the recent change to allow a local directory is a good start towards this

PR #1040, opened 2022-11-17
Fixing `load adcirc` due to lost of `ADCIRCDIR` (put it back).

### LINKED ISSUE #1038: Issue 1004 merge breaks `  load adcirc`
Changed `ADCIRCDIR` to ADCIRC_WORK_DIR` , changing back.

wwlwpd:
Issue 1038: The changes in #1004 took renamed ADCIRCDIR to ADCIRC_WORK_DIR, but this had effects outside of `bin/init-adcirc.sh` (which does the building and contains the wizard). This should not have been done since `ADCIRCDIR` is used externally, e.g., in `./asgs_main.sh`; not to mention the most obvious place that spawned this case, in `load adcirc`. So rather than changing `ADCIRCDIR` everwhere, it was restored. `ADCIRCDIR` preceeded `bin/init-adcirc.sh`.

Resolves #1038.

### DIFF
--- a/bin/init-adcirc.sh
+++ b/bin/init-adcirc.sh
@@ -215,7 +215,7 @@ if [[ -d "${ADCIRCBASE}" ]]; then
 fi
 
 # some variables based on ADCIRCBASE that we can define now
-ADCIRC_WORK_DIR=${ADCIRCBASE}/work
+ADCIRCDIR=${ADCIRCBASE}/work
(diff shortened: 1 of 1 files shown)


wwlwpd:
verified to work by @wwlwpd and @GStefK

GStefK (approved):

PR #1055, opened 2022-12-08
Correctly process empty station files

### LINKED ISSUE #1054: Make `control_file_gen.pl` handle empty station files properly
When including interpolation station lists (e.g., `ELEVSTATIONS`) into the `fort.15` file, the ASGS expects `null` as the station file name if there are no stations, and if the station file name is not `null`, then it expects to read a list of stations from the specified file and to count them. If the station file name is provided but is empty, then it correctly counts the number of stations as zero but also ends up erroneously adding an extra line to the `fort.15` file that causes parsing errors when it is read by the code itself. 

Instead, `control_file_gen.pl` should detect that the station file exists but is empty and prevent the erroneous extra line from being written to the `fort.15`.

jasonfleming:
Resolves #1054

Also handles station files with empty lines (or lines that consist only of spaces). Successfully tested.

### DIFF
--- a/control_file_gen.pl
+++ b/control_file_gen.pl
@@ -855,6 +855,7 @@ sub getStations {
    while (<STATIONS>) {
       $_ =~ s/^\s+//;
       next if (substr($_,0,1) eq '#');  # skip comment lines in the stations file
+      next if ($_ eq '');               # skip empty lines in the stations file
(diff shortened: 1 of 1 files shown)


wwlwpd (approved):
Looks good!

PR #1087, opened 2023-02-22
Added `scp` transport for downloading static input

### LINKED ISSUE #1082: add `scp` as a transport method for static file URLs

jasonfleming:
resolves #1082 and I included some explanatory comments ... cleaned up the existing logic as well ... tested the new code with `scp` from `fortytwo` using adcirc-global-test mesh ... also successfully tested modified version with existing digitalocean cloud cdn using `curl` to confirm that it still works.

### DIFF
--- a/asgs_main.sh
+++ b/asgs_main.sh
@@ -117,7 +117,7 @@ checkFileExistence()
         local downloadCMD
         if [[ $URL =~ "http://" || $URL =~ "https://" ]]; then
            logMessage "$THIS: The curl version is $(curl --version)"
-           downloadCMD="curl ${URL}/${FNAME}.xz --output ${FPATH}/${FNAME}.xz"
+           downloadCMD="curl --insecure ${URL}/${FNAME}.xz --output ${FPATH}/${FNAME}.xz"
         elif [[ $URL =~ "scp://" ]]; then
            downloadCMD="scp $URL/${FNAME}.xz $FPATH/${FNAME}.xz"
         else
--- a/asgs_main.sh
+++ b/asgs_main.sh
@@ -75,34 +75,65 @@ checkFileExistence()
   if [[ -z $FNAME ]]; then
      fatal "$THIS: The $FTYPE was not specified in the configuration file. When it is specified, the ASGS will look for it in the path ${FPATH}."
   fi
-  success=no
+  local success=no
   if [ $FNAME ]; then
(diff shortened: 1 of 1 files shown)


asgs_main.sh
```diff
@@ -75,34 +75,65 @@ checkFileExistence()
...
+           *)
+              warn "$THIS: Unrecognized file type to download: '$FTYPE'."
+              URL="unknown"
+              ;;
+        esac
+        local downloadCMD
+        if [[ $URL =~ "http://" || $URL =~ "https://" ]]; then
+           logMessage "$THIS: The curl version is $(curl --version)"
+           downloadCMD="curl ${URL}/${FNAME}.xz --output ${FPATH}/${FNAME}.xz"
+        elif [[ $URL =~ "scp://" ]]; then
```
wwlwpd, line 121: Sent you a DM about this one, `scp://` may not be what we want. And if it is this is not looking sufficient to handle it like how you mean here (via `scp`).

wwlwpd (changes requested):
Minor code suggestion, but moreover we should discuss the `scp://` URI scheme.

asgs_main.sh
```diff
@@ -75,34 +75,65 @@ checkFileExistence()
...
+              ;;
+           *)
+              warn "$THIS: Unrecognized file type to download: '$FTYPE'."
+              URL="unknown"
+              ;;
+        esac
+        local downloadCMD
+        if [[ $URL =~ "http://" || $URL =~ "https://" ]]; then
+           logMessage "$THIS: The curl version is $(curl --version)"
+           downloadCMD="curl ${URL}/${FNAME}.xz --output ${FPATH}/${FNAME}.xz"
```
wwlwpd, line 120: We may want to turn off SSL host validation here or provide a way to do it (i.e., pass `curl` the `

jasonfleming: Ok I've fixed this by adding `--insecure` thank you

wwlwpd:
Turns out OpenSSH has supported `scp://...` since version 7.7:

asgs_main.sh
wwlwpd: See my latest message, `scp://` seems legit.

wwlwpd (dismissed):
Provide this passes your testing, I approve it.

Please move this out of draft when you're ready, and I will approve it again and merge it.

wwlwpd (approved):
looks good, great advancement!

Issue #1002, opened 2022-09-17
upgrade URL hotstart capability to use either `fort.68.nc` or `fort.67.nc`


jasonfleming:
When downloading a hotstart file from a URL, `asgs_main.sh` looks for a `fort.68.nc` file only. However, only a `fort.67.nc` hotstart file is generated by initialization scenarios, and nowcast scenarios post both `fort.67.nc` and `fort.68.nc` files. At a minimum, we need to upgrade so that hotstarting from a `fort.67.nc` file from a URL is supported, and ideally Operator configurable.

PR #1088, opened 2023-02-22
Add `scp://`, `ssh://` transports

### LINKED ISSUE #715: add capability to hotstart via scp the way we currently do with http
It would enhance operational flexibility if the ASGS could hotstart by downloading a hotstart file and run.properties file via scp the way we currently do with http. This is helpful in situations where the fault in the remote system is with the posting to thredds or an issue with the thredds server itself or the system it is running on.

jasonfleming:
resolves #715 not tested yet

(edited by @wwlwpd below)

We discussed on 4/9/24, updating this PR:

- [x] support both `scp://` and `ssh://` due to the upstream desire to deprecated `scp`, but also user preference (it is okay if `ssh://` uses `scp` underneath
- [x] ~~if it looks like a file (i.e., doesn't have `http://`, `https://`, or `ssh://`), look for the standard `HOST:/path/to/file`; this can be done easily by splitting on the `:`, with `awk -F: '{print $1}'` and `awk -F '{print $2}'`.~~ - I suggest we strike this
- [x] These things to do will round out this support nicely. In the future, other protocols may become important (e.g., `s3://`, etc); so we definitely want to be able to have a clear place to support them and a consistent way to support them across the ASGS toolset.
- [x] add `INITINPUT` variable that will run a specified script, if defined, at input fetch time
- [x] update CERA mesh_defaults.sh with `INITINPUT` ([see](https://github.com/CERA-GROUP/asgs-local-assets/pull/3))
- [x] update docs ([wiki](https://github.com/StormSurgeLive/asgs/wiki/Adding-a-New-Mesh), in mesh_default.sh files)

### DIFF
--- a/asgs_main.sh
+++ b/asgs_main.sh
@@ -1806,16 +1806,24 @@ hotstartBase=fort.${LUN}
 hotstartSuffix=.nc
 hotstartPath=${LASTSUBDIR}/nowcast # only for reading from local filesystem
 hotstartURL=null
+hotstartDownloadExecutable="curl"
+hotstartDownloadRedirect="--output"
 if [[ $HOTORCOLD = hotstart ]]; then
    consoleMessage "$I Acquiring hotstart file."
    # check to see if the LASTSUBDIR is actually a URL
-   urlCheck=$(expr match "$LASTSUBDIR" 'http')
-   if [[ $urlCheck -eq 4 ]]; then
+   if [[ $LASTSUBDIR =~ "http://" || $LASTSUBDIR =~ "https://" || $LASTSUBDIR =~ "scp://" ]]; then
       # always look for fort.68.nc from a URL because only a forecast
       # will be posted to a URL, and only the hotstart file that was used
       # to start the forecast will be posted ... asgs always hotstarts from
       # a fort.68 file and always writes a fort.67 file
       hotstartURL=$LASTSUBDIR
+      # If scp is to be used, the host where the hotstart (and run.properties) files
+      # are downloaded from must support public key authentication. The URL is expected
+      # to be in the form scp://tacc_tds3//full/path/to/hotstart/file
+      if [[ $LASTSUBDIR =~ "scp://" ]]; then
+         hotstartDownloadExecutable="scp"
+         hotstartDownloadRedirect=""
+      fi
    else
       # we are reading the hotstart file from the local filesystem, determine
       # whether it is from a nowcast or hindcast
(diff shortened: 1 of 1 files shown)


wwlwpd (approved):
If you want to take this out of _draft_ and confirm for me it was working for you last time you used it, I'd be happy to merge it if you're ready for it.

wwlwpd:
Thanks @jasonfleming - bear in mind if we just want to support it in this particular case, the I think the current implementation is fine. If we want to add `scp://` support in other places, we should create a function that does parsing of the path so we're not re-implementing that all over the place. I am happy to help here.

wwlwpd:
@jasonfleming just wondering what the state of this PR draft was

jasonfleming:
Implemented but not tested. I just bumped the priority of the underlying issue to important non-critical

wwlwpd:
We can discuss this after #1277 gets in. I've opted to not use a URI scheme for now, instead assuming `ssh` is the transport, and just using the direct syntax used by `ssh`, `scp`, and `rsync` (which assumes `ssh`), `user@host:path/to/file`.

more details:

Since we lean heavily on `~/.ssh/config`, it's more practically, `host:path/to/file`. This is both easier to parse and can be used directly during `backup`. For `restore`, I just `awk -F:` to get `host` and `file` since the command to get the files using `ssh` necessarily uses the parts in a deconstructed way. See #1277 for more context to the snippets below,

```
BKPTGZ=${1:-$HOME/ASGS-SAVE-${PWD}.tgz
#... looks for ":" to indicate remote file path
if [[ $BKPTGZ =~ ":" ]]; then
  USE_SSH=yes
fi
```
...
```
if [ "$USE_SSH" == "yes" ]; then
  HOST=$(echo "$BKPTGZ"|awk -F: '{print $1}')
  FILE=$(echo "$BKPTGZ"|awk -F: '{print $2}')
  tar zcvf - -T ./ASGS-MANIFEST | ssh $HOST "cat > $FILE"
else         
  tar zcvf $BKPTGZ -T ./ASGS-MANIFEST
fi
```

jasonfleming:
The usual transport for hotstart files is https ... maybe if the ":" is missing then we assume the transport is https?

jasonfleming:
Need to include a fix for #1002 for this to be useful.

wwlwpd:
@jasonfleming I have a few things left to do, but you can do an initial review here. I recommend we only test what we normally use. e.g., eye-ball the `gz` support but don't worry about testing it. Similarly, do not worry about testing the (upcoming) `INPUTINIT` script support I am adding to support @enikidis's `asgs-lfs.sh` script.

I will need you to closely look or test if you have time that the URL schemes are working for the original intent of #715, which was to allow hotstart files to be specified via `scp://` or `ssh://` URIs (like we can do with `http`).

wwlwpd:
@jasonfleming - Documentation could be an endless process, I stub'd out all the major parts, I think. Anything major that's obviously missing? We should develop some of the smaller sections at some point - but I am shooting for "good enough to remind **us**" :-) - what do you think? https://github.com/StormSurgeLive/asgs/wiki/Adding-a-New-Mesh

wwlwpd (commented):
@jasonfleming I need you to review this, but I am not sure how to get GH to let me add you. It has me as the one who needs to review it lol. TY.

wwlwpd:
I am going to recreate teh PR.

Issue #1114, opened 2023-05-02
support better `asgs-sendmail` troubleshooting


wwlwpd:
- [x] add `$SCRIPTDIR/mail.log` support to `inspect` and  `edit` commands
- [x] add `$HOME/.asgsh_profile` support to `inspect` and  `edit` commands
- [x] add `$HOME/asgs-global.conf` support to `inspect` and  `edit` commands
- [x] add an interactive `verify sendmail` command
- [x] add `mail.log` support as `tailf mail-log`, update bash completion for `tailf`

PR #1118, opened 2023-05-05
Fixed asgs-sendmail to warn on long subject line and truncate.

wwlwpd:
Issue 1114: Additional commit to this issue, a fix to recently discovered sensitivity to subject lengths > 78. asgs-sendmail will now warn when the subject is > 78 characters, and in an attempt to "just work", truncates the subject and sends it anyway.

Related to #1114.

Note: Change includes fixes to some nagging warnings I observed during the pre-drill forecasts I was running.

### DIFF
--- a/bin/asgs-sendmail
+++ b/bin/asgs-sendmail
@@ -25,22 +25,23 @@ package bin::asgs_sendmail;
 use strict;
 use warnings;
 
-use Email::Sender::Simple qw(sendmail);  # main email capability (receives TLS obj)
-use Email::Sender::Transport::SMTP::TLS; # creates transport object
-use Email::Simple::Creator;  # generates message itself (converts header and body from hash to text email)
-use Config::Tiny; # reads ini format ... converts to a perl hash
-use Try::Tiny;    # exception handling (alternative to eval)
+use Email::Sender::Simple qw(sendmail);     # main email capability (receives TLS obj)
+use Email::Sender::Transport::SMTP::TLS;    # creates transport object
+use Email::Simple::Creator;                 # generates message itself (converts header and body from hash to text email)
+use Config::Tiny;                           # reads ini format ... converts to a perl hash
+use Try::Tiny;                              # exception handling (alternative to eval)
 use POSIX qw/strftime/;
 use Getopt::Long;
-use Util::H2O qw/h2o/;
+use Util::H2O::More qw/h2o ini2h2o/;
 use constant EXIT_SUCCESS => 0;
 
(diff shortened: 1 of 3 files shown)


bin/asgs-sendmail
```diff
@@ -76,34 +78,40 @@ sub run {
...
     catch {
-      logger($logfile, qq{(FAILED) $logmsg - $_});
-      logger( type => q{FAILED}, file => $logfile, msg => qq{$logmsg - $_});
-      die "Error sending email: $_";
+        logger( $logfile, qq{(FAILED) $logmsg - $_} );
+        logger( type => q{FAILED}, file => $logfile, msg => qq{$logmsg - $_} );
+        die "Error sending email: $_";
     };
 
+    # check for subjects > 78 characters, which is the max line length according to the
```
wwlwpd, line 104: @jasonfleming this is the effective change, I think this is what was causing the weird issues in email in a lot of cases. If the subject line is > 78 characters, the email just doesn't make it to where it is going. I wonder how long this has been happening!

bin/asgs-sendmail
```diff
@@ -76,34 +78,40 @@ sub run {
...
+    # email RFC
+    if ( length $subject > 78 ) {
+        warn qq{(asgs-sendmail) WARNING: Length of --subject is > 78 characters, truncating ...\n};
+        $subject = substr $subject, 0, 78;
+    }
+
```
wwlwpd, line 110: @jasonfleming **above** is comment is the effective change, I think this is what was causing the weird issues in email in a lot of cases. If the subject line is > 78 characters, the email just doesn't make it to where it is going. I wonder how long this has been happening!

jasonfleming: I like this change because the email still gets through and unless it breaks on a space, it will be obvious that it is truncated, which adds to the visibility of the issue. There is a limit to the metadata we can embed in the subject line.

wwlwpd:
Testing successfully. Details: I've got 2 different ASGSes running on different machines, and emails are coming through just fine. Also, one of them has a rather long `INSTANCENAME` and `ASGSADMIN`; so I don't think this is going to impact either too greatly in practice.

An important constraint affected is that a unique identifier is positioned at the very end (`ASGSADMIN`). Most impacted element will be `INSTANCENAME`. I've proposed we wait to address any change in the format of the subject wrt downstream processing until after the 2023 hurricane season.

wwlwpd:
Closing, been superceded by #1126.

PR #1131, opened 2023-05-13
Made call to "checkTimeLimit" in asgs_main.sh more robust.

### LINKED ISSUE #1130: `asgs_main.sh`: `checkTimeLimit` call evaluating to  syntax error
The current line evaluates to a syntax error when `$(checkTimeLimit "$startTime" "$WALLTIME")` is empty, at:

```
if [[ $(checkTimeLimit "$startTime" "$WALLTIME") -eq 1 ]]; then
```

Effective evalues to this, which is a syntax error:

```
if [[ -eq 1 ]]; then 
```

wwlwpd:
Issue 1130:

The old line evaluated to a syntax error when $(checkTimeLimit "$startTime" "$WALLTIME") is empty, at:

  if [[ $(checkTimeLimit "$startTime" "$WALLTIME") -eq 1 ]]; then

Effectively evalued to this, which is a syntax error:

  if [[ -eq 1 ]]; then

This change doesn't address why the valuation occurs, which would be more of a root issue; however the situation is now robustly detected and reported via the logging function, "warn".

Resolves #1130.

### DIFF
--- a/asgs_main.sh
+++ b/asgs_main.sh
@@ -1054,9 +1054,13 @@ downloadRiverFluxData()
 # See if a task has been running (or waiting) longer than a specified time limit
 checkTimeLimit()
 {
+   THIS="asgs_main.sh>checkTimeLimit()"
+   if [[ -z "$1" || -z "$2" ]]; then
+	   warn "$ENSTORM: $THIS: One or both parameters for checkTimeLimit() is empty. STARTTIME='$1', TIMELIMIT='$2'."
+     return 0
+   fi
    STARTTIME=$1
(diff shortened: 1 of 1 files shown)


jasonfleming:
I like this but I am concerned that empty variables are being passed to `checkTimeLimit`. This fix will help us troubleshoot to determine root cause.

jasonfleming (approved):
This looks good and will give us insight into what is going on with these empty variables.

jasonfleming (approved):
This looks good and will give us insight into what is going on with these empty variables.

PR #1186, opened 2023-08-21
Added a queue hint for mike III when using up to 1 full node.

### LINKED ISSUE #1185: Use single queue on mike3 when submitting parallel jobs on less than 64 cores

wwlwpd:
Issue 1185: LSU HPC now requires the 'single' queue to be used when using single compute node up to the full 64 processors.

Resolves #1185.

### DIFF
--- a/platforms.sh
+++ b/platforms.sh
@@ -508,6 +508,15 @@ HPC_Queue_Hint()
        echo $DEFAULT_QUEUENAME
      fi
    ;;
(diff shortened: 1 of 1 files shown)


jasonfleming:
I had a look, it looks good to me, I will approve it as soon as you confirm that it works as expected.

jasonfleming (approved):
Looks great. I will put this into production immediately. :-)

PR #1198, opened 2023-09-23
Ensure STATEFILE file name generated from INSTANCENAME is valid

### LINKED ISSUE #1196: Allow `INSTANCENAME` to be enclosed in quotes in config file without mangling `STATEFILE` name
I recently found that when the `INSTANCENAME` in the ASGS config file is surrounded by double quotes, this somehow contaminates the name of the `STATEFILE` in the ASGS profile as follows:
```bash
[ASGS (master)] frontera@production> load profile HSOFS_gfs_jgf
(info) loaded 'HSOFS_gfs_jgf' into current profile
(info) config file found, instance name is '"HSOFS_gfs_jgf"'

(!! warning) state file '/scratch1/00976/jgflemin/"HSOFS_gfs_jgf".state' does not exist.

(info) no indication of first run yet?
```
The actual `STATEFILE` was there and was named HSOFS_gfs_jgf.state as expected. Not sure whether fixing this would require a change in the ASGS Shell or if it should be added to the linter.

wwlwpd:
Issue 1196: Added a sed command to ensure that any filename made for use as STATEFILE from INSTANCENAME will not contain invalid characters.

Resolves #1196.

### DIFF
--- a/cloud/general/DOT-asgs-brew.sh
+++ b/cloud/general/DOT-asgs-brew.sh
@@ -314,7 +314,7 @@ _parse_config() {
     return
   fi
   # pull out var info the old fashion way...
-  export INSTANCENAME=$(egrep '^ *INSTANCENAME=' "${1}" | sed 's/^ *INSTANCENAME=//' | sed 's/ *#.*$//g')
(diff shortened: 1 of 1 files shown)


wwlwpd:
E.g., 

```
INSTANCENAME="CPRA23v01c_nam_be"  # "name" of this ASGS process
```

results in:

```
(!! warning) state file '/work/asgs/_CPRA23v01c_nam_be_.state' does not exist.
```

PR #1209, opened 2023-10-06
Began cleaning up operators view of running ASGS.

wwlwpd:
@jasonfleming - let me know what you think so far:

Sample of what it looks like now:

image


wwlwpd:
I think this is not exactly what we want to be showing from an operational POV, but it cleans up the spew quite a bit from `opendap_post2.sh`.

wwlwpd:
@jasonfleming I am going to be putting this into a real test with the daily on qbC

wwlwpd:
image

monitoring/logging.sh
```diff
@@ -76,7 +76,7 @@ finalizeCentralizedScenarioLogging() {
    for pid in ${subshellPIDs[*]}; do
       # terminate each one
       echo "Terminating previously spawned subshell process ID ${pid}." | awk -v level=INFO -v this=logging.sh -f $SCRIPTDIR/monitoring/timestamp.awk >> scenario.log 2>&1
-      kill -TERM $pid 2>&1 | awk -v this=logging.sh -v level=INFO -f $SCRIPTDIR/monitoring/timestamp.awk >> scenario.log 2>&1
+      kill  ERM $pid 2>&1 | awk -v this=logging.sh -v level=INFO -f $SCRIPTDIR/monitoring/timestamp.awk >> scenario.log 2>&1
```
jasonfleming, line 79: Is this a typo?

monitoring/logging.sh
```diff
@@ -99,21 +99,22 @@ findAndClearOrphans() {
...
-      echo "[$DATETIME] INFO: Created subdirectory for ASGS log files : '$WORK/log'."
+      echo "[$DATETIME] Created subdirectory for ASGS log files : '$WORK/log'."
    else
-      echo "[$DATETIME] INFO: Found subdirectory for ASGS log files : '$WORK/log'."
+      echo "[$DATETIME] Found subdirectory for ASGS log files : '$WORK/log'."
    fi
    SYSLOG=${SYSLOG:-$WORK/log/${INSTANCENAME}.asgs-${STARTDATETIME}.$$.log}
-   consoleMessage "Set ASGS log file parameter SYSLOG to '$SYSLOG'." $SYSLOG
+   consoleMessage "SYSLOG set to"
+   consoleMessage "$SYSLOG'." $SYSLOG
```
jasonfleming, line 111: why are these separate messages?

monitoring/logging.sh
```diff
@@ -184,15 +185,15 @@ allMessage()
...
+{ local DATETIME=`date +'%D %H:%M:%S'`
   MSG="[${DATETIME}] WARNING: $1"
   for warnlogfile in $SYSLOG $2 ; do
     if [[ -e $warnlogfile ]]; then
       local frame=0
-      while caller $frame; do
-         caller $frame >> $warnlogfile
-         ((frame++));
-      done
+ #     while caller $frame; do
```
jasonfleming, line 193: I find the execution tracing on warning messages to be very helpful ... is there a better way for us to be doing this?

jasonfleming:
I like the idea of making the date/time stamp on our log messages consistent across the board ... can we do that as a separate PR or set of related PRs?

monitoring/logging.sh
wwlwpd: I did that because `$SYSLOG` can be a very long path. I can look at how to do that without invoking `consoleMessage` twice like that.

monitoring/logging.sh
wwlwpd: This was causing uncatchable chatter in the main window. I can look more into that call and see what options we have.

monitoring/logging.sh
wwlwpd: Yes! Good catch.

wwlwpd:
> I like the idea of making the date/time stamp on our log messages consistent across the board ... can we do that as a separate PR or set of related PRs?

Probably just include it all - I think that's the net result of most of the chances reflected here.

monitoring/logging.sh
wwlwpd: fixed

monitoring/logging.sh
wwlwpd: FIXED

monitoring/logging.sh
wwlwpd: I looked more into this. I understand now `caller` is to create basically a stack trace in the event of an error. I'll uncomment this, but we'll want to fix the underlying issue that is actually calling `warn`.

jasonfleming: That sounds great, I am open to better solutions.

jasonfleming:
Let's talk about next steps on this in our next meeting. There is a lot of room for improvement in our UX and this is a big step in the right direction.

PR #1212, opened 2023-10-12
Enabled Operators to skip creation of fulldomain SWAN hotstart file after nowcast

### LINKED ISSUE #1210: Allow Operators to turn off the recomposition of SWAN hotstart files after a nowcast via config file
The capability to recompose and decompose fulldomain SWAN hotstart files from subdomain SWAN hotstart files is less commonly used than in the past. The utility that performs this task also runs in serial, takes a long time, and uses a lot of memory. As a result, in many cases Operators would benefit from being able to turn off this step of nowcast post processing.

jasonfleming:
resolves #1210


wwlwpd (approved):
looks good to me

PR #1255, opened 2024-01-18
Removed confusing dialog about checking out branch during "build adci…

### LINKED ISSUE #1254: in `build adcirc` wizard, make it more clear how to skip the checkout of a git rep
image

* `'.' to skip` should just be something like, "_hit <enter> if you don't know_"
* is the _skip_ option even needed? for debugging/troubleshooting?

wwlwpd:
…rc".

Issue 1254: this dialog was needed in the early days, but now only adds confusion for users.

Resolves #1254.


jasonfleming (approved):
Nice simplification. Looks ready to merge. :-)

PR #1260, opened 2024-02-07
Changes to support model parameter modification in asgs config

### LINKED ISSUE #255: enable inundation output from adcirc via asgs config
Also create a standard setting (on by default) in mesh_defaults.sh. This could be implemented with a %inundationoutput% line at the end of each fort.15 template and by teaching control_file_gen.pl to fill it in as specified.

### LINKED ISSUE #920: remove absolute path reference to `tide_fac.x` from `control_file_gen.pl`
(Note: I ran into this debugging an `asgs` install that had gotten corrupted somehow, but would be good to fix nonetheless).

`tide_fac.x` is in `PATH` so this shouldn't be required any more.

```
my $tides = "off";
if ( -e "$scriptdir/tides/tide_fac.x" && -x "$scriptdir/tides/tide_fac.x" ) {
   my $tide_fac_message = `$scriptdir/tides/tide_fac.x --length $RNDAY --year $cy --month $cm --day $cd --hour $ch -n 8 m2 s2 n2 k1 k2 o1 q1 p1 --outputformat simple --outputdir $stormDir`;
   if ( $tide_fac_message =~ /ERROR|WARNING/ ) {
      stderrMessage("WARNING","There was an issue when running $scriptdir/tides/tide_fac.x: $tide_fac_message.");
   } else {
```

Also, there's a `sleep 60` in the error handler that probably should be removed; but I need to find out it's intended purpose first. @jasonfleming

### LINKED ISSUE #1244: extend configurability to parameter packages
Additional functionality is required for parameter configuration to be achieved through the ASGS configuration.sh script. Two examples are the `sea_surface_height_above_geoid` (fort.13) and the wind drag parameter (fort.15). Our goal would be to be able to set these in the config.sh script for an individual run. ASGS would then pick up those values and overwrite the default values as given in the input files/mesh_defaults.sh or any other input.

In order to implement this, we will add the concept of parameter packages in a way that is similar to the way we currently handle scenario packages (e.g., veerRight100, veerLeft100, etc). The different combinations of parameters will each have a name, and the parameter values will be set in the `config.sh` based on the name of the parameter package. These packages can also have a number associated with them so that they can be iterated over. 

New settings will have to be added to the `mesh_defaults.sh` script in order to control these values. We will also need to update `control_file_gen.pl` so that it can fill in the fort.15 template with different parameters; we will need to create fort.13 template files that allow Operators to fill in values dynamically; we also need to generate metadata dynamically based on these settings for transmission in `properties` files.

### LINKED ISSUE #1253: Update `fort.15` template to support linting, debugging, and troubleshooting
Because of the "unique" nature of the ADCIRC `fort.15` file format, it can be very difficult to detect errors or to determine what has gone wrong after an error is reported by `adcirc` or `adcprep`. In order to support linting, debugging, and troubleshooting this file, the ASGS `fort.15` templates can be upgraded with hints. These hints will be used to proactively and reactively diagnose issues and errors.

jasonfleming:
resolves #1244 resolves #1253 resolves #255 resolves #920


jasonfleming:
@wwlwpd I have tested this successfully and it is ready for review.

wwlwpd (approved):
Big change, but looks good!!

PR #1263, opened 2024-02-17
Added `TRIGGER=auto` as a config option for vortex model hindcasts to eliminate a need to generate mock xml forecasts

### LINKED ISSUE #1247: enable nowcast initiation without providing a mock `index-at.xml`
When hindcasting for historical storms using a vortex model, only the BEST track file is really required but the ASGS uses an `index-at.xml` file to know when to kick off a nowcast/forecast cycle. As a result, for hindcasts, Operators are still required to provide a mock `index-at.xml` to let the ASGS know that new data are available to trigger the next cycle. This requirement should be removed, so that Operators can provide a BEST track file and perform a hindcast without needing to provide a mock forecast/advisory.

jasonfleming:
resolves #1247

### DIFF
--- a/asgs_main.sh
+++ b/asgs_main.sh
@@ -3274,7 +3274,6 @@ while [ true ]; do
       OLDADVISDIR=$ADVISDIR
    fi
    if [[ $ONESHOT = yes || $NOFORCING = true || $TRIGGER = auto ]]; then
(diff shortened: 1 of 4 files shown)


wwlwpd (approved):
Looks great to me, and I know someone who will really appreciate this change <grin>. @enikidis

Issue #1258, opened 2024-02-05
settings in `config/io_defaults.sh` may not work for combination of netCDF3 hotstart files and Wind10m layers


jasonfleming:
Have discovered that `fort.15` files generated to trigger met-only mode are actually still trying to write netCDF3 hotstart files if the scenario was configured to do so, thus deactivating met-only mode. Need to reproduce and then fix the `config/io_defaults.sh` settings.

jasonfleming:
Resolved by #1266

PR #1266, opened 2024-02-20
Integrated generation of wind10m layer into workflow to eliminate need for separate scenario

### LINKED ISSUE #1251: Enable inline production of Wind10m layer rather than producing it as a separate scenario
We originally implemented the production of the Wind10m layer as a separate forecast scenario as a short term workaround. However, over time, this layer has grown in importance (including for nowcasts) and its implementation as a separate forecast scenario that must be manually configured is suboptimal for many reasons. As a result, some refactoring of `asgs_main.sh` and the queue script template should be performed to enable the production of this layer by default. The Operator will have the option of turning off the Wind10m layer if that is preferred.

jasonfleming:
resolves #1251

### DIFF
--- a/config/mesh_defaults.sh
+++ b/config/mesh_defaults.sh
@@ -64,7 +64,7 @@ CONTROLTEMPLATENOROUGH="null"
 #
 #
 case $GRIDNAME in
-   "CPRA23v05b")
+   "CPRA23v05b"|"CPRA23v05c")
       #
       nodes=1577268
       elements=3072131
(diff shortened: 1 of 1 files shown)


jasonfleming:
Test plan:

- createWind10mLayer: `yes|no`
- SCENARIO: `*Wind10m|not` 
- nodal_attribute_activate: `empty|list`
- NAFILE: `null|populated`

jasonfleming:
In order to accommodate backward compatibility requirements for pre-existing ASGS config files, I developed 32 test cases for this using many different combinations of specifications, both old and new, and created a meta--automation system to run them. Worked out really well. All of the 32 tests use GFS. Will run 2 hindcast test cases with GAHM also before marking this as ready for review.

jasonfleming:
resolves #1258

wwlwpd (approved):
wow, big change! Looks great, thank you!

PR #1270, opened 2024-03-18
Fixing XDMF variable naming issues and removing vector magnitude

### LINKED ISSUE #1020: Use CF std name instead of netCDF var name in generateXDMF.f90
When writing XML for time varying output, the `Attribute` name should be the standard CF name rather than the name of the corresponding variable in the netCDF file.

### LINKED ISSUE #1053: Make `generateXDMF.f90` process velocity vector quantities in hotstart file correctly
When reading a hotstart file, the `generateXDMF.f90` utility finds and processes the velocity vectors correctly but does not write them to the `.xmf` file correctly. It also does not write out a function for computing vector magnitude.

### LINKED ISSUE #1169: Improve swan max file support in `generateXDMF.f90`
It appears that this utility does not recognize `swan_DIR_max.63.nc`, `swan_TM01_max.63.nc`, `swan_TM02_max.63.nc`, or `swan_TMM10_max.63.nc`.

### LINKED ISSUE #1256: Found an issue in the XDMF Reader for Paraview for SQRT function
Will try replacing SQRT function with EXP raising to the 0.5 power. If that does not work, this code will have to be removed from `generateXDMF.f90`.

jasonfleming:
resolves #1256 resolves #1169 resolves #1053 resolves #1020

### DIFF
--- a/output/asgsio.f90
+++ b/output/asgsio.f90
@@ -690,6 +690,12 @@ do i=1,f%nvar
       f % fileTypeDesc = "a SWAN wave direction (swan_DIR.63) file"
       call initFileMetaData(f, thisVarName, 1, 1)
       exit
+   case("swan_DIR_max")
+      f%defaultFileName = 'swan_DIR_max.63'
+      f % dataFileCategory = MINMAX
+      f % fileTypeDesc = "a SWAN wave direction at maximum (swan_DIR_max.63) file"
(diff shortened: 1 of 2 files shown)


jasonfleming:
@wwlwpd I've tested this, and found that it works. Ready for review.

wwlwpd (approved):
Looks great to me.

PR #1290, opened 2024-04-09
Fix for NAM OWI header parsing as well as setting `COLDSTARTDATE`

### LINKED ISSUE #1288: Fix typo in parsing of OWI Win/Pre ASCII file headers for NAM
The field specifications for the time limits of the data contained in the OWI file header are not being parsed correctly.

### LINKED ISSUE #1289: Fix setting of `COLDSTARTDATE` where it is dynamically set in the ASGS config file during hotstart initialization
The code in `asgs_main.sh` that establishes the value of `CSDATE` (datetime corresponding to model cold start) erroneously assumes that the `COLDSTARTDATE` parameter is static in the ASGS config file and set to either a 10-digit datetime or the string "auto" which indicates that the value should be extracted from the `run.properties` file associated with the hotstart file it is starting from. 

However, if the `get-coldstart-date` utility is used inline to set COLDSTARTDATE in the ASGS config file, the value will be dynamic but not "auto" which means that the value of `CSDATE` will be updated each time the ASGS is stopped and restarted, which is not the expected behavior. 

The fix is to always use the cold start datetime associated with the hotstart file whenever starting with a hotstart file, whether or not "auto" has been specified in the ASGS config, and to write a log message if the configured `COLDSTARTDATE` and the value associated with the hotstart file differ.

jasonfleming:
resolves #1288 resolves #1289

### DIFF
--- a/asgs_main.sh
+++ b/asgs_main.sh
@@ -1869,16 +1869,20 @@ if [[ $HOTORCOLD = hotstart ]]; then
       # check to make sure the COLDSTARTDATE was not set to "auto" in the
       # asgs config file (unless the run.properties file was also supplied)
       if [[ $COLDSTARTDATE = auto ]]; then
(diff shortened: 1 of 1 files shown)


wwlwpd (approved):
looks good!

PR #1297, opened 2024-04-17
Added quotes around exported variables in profile for Windows paths.

### LINKED ISSUE #1296: Ubuntu on WSL2 (Windows) causing "export" errors with unquoted Windows style paths
1. happens when `./asgsh` is being written (`PATH`)
2. happens during `save profile` (`ASGS_BREW_FLAGS`)

wwlwpd:
Issue 1296: Windows paths were causing start up issues, and it was discovered that the environmental variables were not being quoted. Additionally, it was discovered and fixed that the variable, "ASGS_BREW_FLAGS", was being quoted too early in the process. This was adjusted to be treated like the others.

In addition to this, the default "$WORK" and "$SCRATCH" directories for "debian" were adjusted so that they wer under $HOME, and not a root level directory that would show unnecessary warnings for non-privileged users.

Resolves #1296.

### DIFF
--- a/cloud/general/DOT-asgs-brew.sh
+++ b/cloud/general/DOT-asgs-brew.sh
@@ -384,7 +384,7 @@ save() {
   # includes an 'export' line for each variable asgsh cares about; this
   # is defined as part of the shell installation by asgs-brew.pl
   for e in $_ASGS_EXPORTED_VARS; do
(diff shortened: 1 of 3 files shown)


jasonfleming (approved):
This looks great thank you. Sorry for the delay.

PR #1319, opened 2024-05-22
Applying the same fix to capture the job ID as was used on mike

### LINKED ISSUE #1318: Fix the json status file for jobs on qbd
This appears to be the same issue reported in #1306 and fixed in #1307

jasonfleming:
Resolves #1318


wwlwpd (approved):
looks good, merging

PR #1330, opened 2024-05-30
Added code to maintain precedence of `OPENDAPNOTIFY` from `run.properties` file

### LINKED ISSUE #1328: Correct the overriding of `OPENDAPNOTIFY` parameter in `output/opendap_post2.sh` when this parameter is specified in `~/.asgsh_profile`
We have recently received email notification of each Wind10m scenario status update from ASGS instances on supermic. The root cause of this behavior (compared to not receiving these emails for other scenario status updates, or on other platforms) is not clear. It is also not clear why we implemented the capability to receive email updates on these very frequent events in the first place (other than perhaps for feedback during initial development). As a result, this particular notification capability will be removed.

jasonfleming:
resolves #1328


jasonfleming:
testing now

jasonfleming:
This is tested and appears to be working.

wwlwpd (approved):
Looks good to me. I know this was a lot of work to find and fix, thank you!

PR #1343, opened 2024-07-01
Fix calculation of vortex forecast end time 

### LINKED ISSUE #1342: Correct the calculation of vortex forecast end time
In `asgs_main.sh` there is a typo that causes the ASGS to use the end date of the BEST track file to calculate forecast duration rather than the end date of the forecast data.

jasonfleming:
resolves #1342


wwlwpd (approved):
Looks good!

PR #1347, opened 2024-07-03
`clone profile` no longer modifies parent profile (ASGS_CONFIG)

### LINKED ISSUE #1346: `clone profile` inadvertently updates `ASGS_CONFIG` in the parent (cloned) profile

wwlwpd:
Issue 1346:

Resolves #1346.

### DIFF
--- a/cloud/general/DOT-asgs-brew.sh
+++ b/cloud/general/DOT-asgs-brew.sh
@@ -464,9 +464,14 @@ clone() {
       fi
       read -p "Create new profile? [y] " create
       if [[ -z "$create" || "$create" = "y"  ]]; then
+        local DIR=$(dirname "${new_config}")
+        if [[ ! -d "$DIR" ]]; then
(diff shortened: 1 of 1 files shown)


wwlwpd:
I was able to reproduce the issue and test the fix. It works now, I added a check also to make sure the potentially new directory for the new configuration exists before proceeding.

jasonfleming (approved):
Looks good may also want to consider the following situations: What if the $new_config already exists? What if the new profile name is the same as an existing profile name?

wwlwpd:
> Looks good may also want to consider the following situations: What if the $new_config already exists? What if the new profile name is the same as an existing profile name?

Good thoughts, as not to add to technical debt I think I am going to merge this. I use this command often in the heat of the moment so if I find that it is biting me I'll revisit. Please let me know if you run into that situation also. (also I want to use this now :-))

PR #1351, opened 2024-07-17
Fixes to support new SuperMIC including switch from PBS to SLURM

### LINKED ISSUE #1334: SuperMIC switching from PBS to SLURM on 15 July

jasonfleming:
Resolves #1334. Lots of unrelated changes on SuperMIC were also encountered, including Intel compiler upgrade from 2019 to 2021 version. Switching the queueing system on an existing HPC machine required very little change in ASGS. In any case, all Operators will have to rebuild from scratch due to the new library versions installed in the underlying OS. 

After our existing ASGS instance has successfully completed a few cycles, I will take this PR out of draft.

### DIFF
--- a/monitoring/captureJobID.sh
+++ b/monitoring/captureJobID.sh
@@ -47,7 +47,7 @@ case $HPCENVSHORT in
    cat jobID.tmp | awk '$1~/[0-9]+/ { print $1 }' > jobID
    rm jobID.tmp
    ;;
-"mike"|"qbd")
+"mike"|"qbd"|"supermic")
    # SLURM returns information similar to the following when a
    # job is submitted:
    # <asgsh> sbatch prep13.slurm 2>stderr >stdout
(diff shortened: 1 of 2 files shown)


wwlwpd (dismissed):
Looks good, I think sq needs to also be updated since it maps a machine to a particular queue command to list any jobs currently in the queue (I use it) - https://github.com/StormSurgeLive/asgs/blob/master/bin/sq

jasonfleming:
Ok good catch. I just checked and found that `showq` continues to work on the new `smic2`. I think it is an LSU thing rather than a PBS thing since the platform lists in the `sq` script are separated according to HPC centers rather by than queueing system. Also we don't have any PBS systems now, so we might get rid of support for `showq` altogether. What do you think?

wwlwpd (approved):
Looks good to.

wwlwpd:
@jasonfleming ready for review? Move it over and I'll do the merge whenever.

wwlwpd:
(still in draft)

jasonfleming:
Thank you! Sorry for the delay. It is running as expected.

PR #1362, opened 2024-08-17
Added code to support edge length calculation and visualization

### LINKED ISSUE #1359: Add netcdf support to `output/checkAdcircMesh.f90`
The `checkAdcircMesh.f90` utility reads an ASCII mesh file and computes various properties and statistics (beyond checking for mesh issues). It would be helpful if this utility could read the mesh from a netcdf file and compute the same properties. Would also be valuable if it could write addition datasets (like mesh resolution) back to the netcdf file for visualization.

jasonfleming:
tested successfully, resolves #1359


wwlwpd:
Issue 1359:

Resolves #1359.
Resolves #1362 (PR).

PR #1376, opened 2024-09-09
Increased timeout for downloading a mesh from 120 sec -> 900 sec.

### LINKED ISSUE #1375: Increase value of timeout for downloading a large mesh file
I was experiencing timeouts on Frontera when downloading a large mesh, then I discovered the timeout was set to 120 seconds. Increasing it to 900 seconds (15 minutes) worked. This may be something we want to make configurable on a per-mesh basis (or make it more of a soft timeout), but for now I think that for now it's fine just set high.

wwlwpd:
Issue 1375: ran into this timeout on Frontera. I adjusted it to 900 seconds, which was just a few minutes over what the mesh needed based on commandline tests. This change worked in this case.

Resolves #1375.

### DIFF
--- a/asgs_main.sh
+++ b/asgs_main.sh
@@ -178,7 +178,7 @@ checkFileExistence()
         consoleMessage "$I Downloading '${FNAME}.xz'"
         $downloadCMD 2> errmsg &
         local pid=$!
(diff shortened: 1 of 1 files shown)


jasonfleming (approved):
Thank you. Not sure why we are seeing these delays, but this will help to prevent the associated issues.

Issue #1386, opened 2024-09-10
update support to  v56 release, others things (super issue)


wwlwpd:
(testing checklist is in PR #1388)

So can rein in the version support here a bit, I propose we pear down primary version support on _all_ platforms to:

- [x] v53.05
- [x] v55.01-5bc04d6
- [x] v55.02
- [x] v56.0.2 (release, https://github.com/adcirc/adcirc/releases/tag/v56.0.2)

In the updated v56 patch supported the latest release,

- [x] remove `-FI` (uppercase F uppercase i) flag on Frontera and Lonestar6 causes `src/netcdf_error.F90` to fail compiling/linking on v56-beta-2 (updated in #1387)
- [x] remove `-FI` on the other platforms that use
- [x] remove v55-release from the list
- [x] address #1361 (gfortran)
- [x] address #1383 (change intel-llvm to intel-openapi, add gfortran >= 10)
- [x] #1311 (Intel OpenAPI support on debian)

wwlwpd:
Note, v56 beta 2 is compiling fine on Stampede3

wwlwpd:
FYI @jasonfleming I think the `-FI` issue only comes into play for v 56+. Based on some cursory research, I believe it can be eliminated for any other version without any ill effects.

PR #1387, opened 2024-09-11
Emergency update to v56 beta 2 patch for Frontera.

wwlwpd:
Issue 1386:

Related to #1386


wwlwpd:
my `bin/repost-all` script snuck in there, which is fine. This is going to be one of potentially several or more PRs for #1386.

wwlwpd:
Tested this on both Frontera and Lonestar6

Issue #1379, opened 2024-09-10
add alias or bash functions for the following tasks during ops ...


wwlwpd:
1. commit and push `$ASGS_CONFIG`,
```
pushd git/asgs-configs             && \
git fetch --all                    && \
git commit -a -m "updating config" && \
git push origin master && popd
```

2. jump into
```
push git/asgs-mon # manually type "popd" to go back to starting point
```

... probably more

PR #1390, opened 2024-09-14
Removed trailing '/' in asgs-mon's PATH entry in asgs-brew

wwlwpd:
Issue 1379: minor thing I noticed when doing, `which asgs-mon`; adjusted "asmv" alias to use fact asgs-mon is in PATH now.

Related to #1379


wwlwpd:
minor things I missed/wanted to change

PR #1395, opened 2024-09-22
Issue with caller capturing value of checkTimeLimit bash function

### LINKED ISSUE #1394: `checkTimeLimit` - line 1199: [[: 1148 checkTimeLimit asgs_main.sh
happening because of the use of `return` versus `echo` - the way it's being called suggests `echo` should be used; when `return` is used, the value is available via `$?`

wwlwpd:
Issue 1394:

Resolve #1394

### DIFF
--- a/asgs_main.sh
+++ b/asgs_main.sh
@@ -938,7 +938,7 @@ prepFile()
 
          fi
       done
-      monitorJobs $QUEUESYS ${JOBTYPE} ${ENSTORM} $WALLTIME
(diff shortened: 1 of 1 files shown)


wwlwpd:
Testing now

wwlwpd:
Working much better, no errors now

wwlwpd:
I made another change, I was still getting some bash errors when time was exceeded

wwlwpd:
This proved helpful to correct log spew seen during Helene on Stampede3 and Lonestart6.

PR #1401, opened 2024-10-06
Updating OpenMPI from 1.8.1 -> 5.0.5

### LINKED ISSUE #1400: update provided OpenMPI (when `mpif90` is not already present) from 1.8.1 -> 5.0.5 (latest)

wwlwpd:
Issue 1400: ASGS attempts to build OpenMPI when it can't find `mpif90`, and for the longest time it was providing an "ancient" version (1.8.1). Recent attempts to support Ubuntu 22 had provided enough motivation to update this version of OpenMPI to the latest current upstream version, which is 5.0.5. It is likely this change will impact older operating systems which needed 1.8.1 to be installed. If this occurs, it should be noted there that ASGS only needs there to be an `mpif90` for compiling ADCIRC. It doesn't care itself how that is provided. But the successful building of ADCIRC does require that the `mpif90` available via PATH be compatible with the C and Fortran compilers that are also being used by everything else.

The reason this upgrade was necessary is because Ubuntu 24.x started providing some development libraries that were no longer compatible with what OpenMPI 1.8.1 was expecting.

Resolves #1400.


wwlwpd:
I tested this on Ubuntu 24.04.1 LTS (currently latest default on Windows WSL2). builds all ADCIRCs we care about, and `verify adcirc` confirmed that MPI works for all of them during run time.

jasonfleming (approved):
This looks great @wwlwpd ... let me know if there is a need to add a file archive for this version of openmpi to our digitalocean space.

wwlwpd:
> This looks great @wwlwpd ... let me know if there is a need to add a file archive for this version of openmpi to our digitalocean space.

I do not think there is a need, but if you wanted to I would be happy to adjust the URL.

PR #1415, opened 2024-11-13
Fixing supermic to work with latest bin/sq via QUEUESYS

### LINKED ISSUE #890: `sq` wrapper around platform's way to check queue status is not compatible with new platform infrastructure
Driving need: add `sq` support for `supermike iii`

THE FIX: use `QUEUESYS` for all

wwlwpd:
Issue 890: supermic is one of the few remaining "old" platforms defined directly in platforms.sh. It (and another) platform didn't export "QUEUESYS" directly, so this was added. The latest bin/sq works by inspecting QUEUESYS in the environment rather than using bin/guess and a case statement on the platform name. This change also cleaned up some remaining artifacts from old platforms, incl. stampede2 and poseidon.

Resolves #890.

### DIFF
--- a/bin/guess
+++ b/bin/guess
@@ -26,8 +26,6 @@ _guess_platform()
     default_platform=supermic
   elif [ 1 -eq $(echo $HOSTNAME | grep -c ls5) ]; then
     default_platform=lonestar5
-  elif [ 1 -eq $(echo $HOSTNAME | grep -c stampede2) ]; then
-    default_platform=stampede2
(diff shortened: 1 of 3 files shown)


jasonfleming (approved):
This looks good! Do I need to test it? If not, please merge. :-)

wwlwpd:
> This looks good! Do I need to test it? If not, please merge. :-)

I tested it working with `sq` - it should not break anything so I think just a visual inspection is sufficient.

PR #1423, opened 2025-01-21
Moved writing of BEST track properties to earlier in the script

### LINKED ISSUE #1419: `TRIGGER=auto` may not work as expected for nowcasting historical storms

### LINKED ISSUE #1425: Calculation of periodic river boundary condition does not account for specification of static flux
Operators can specify a script that will be used to specify the flux at periodic river boundaries. Alternatively, a static flux file can be specified by setting `FLUXCALCULATOR=static`. The latter option was not recognized, and the parameter value `static` was being treated as the name of a script.

jasonfleming:
resolves #1419 resolves #1425

### DIFF
--- a/asgs_main.sh
+++ b/asgs_main.sh
@@ -276,7 +276,7 @@ checkHotstart()
    # composition, and location rather than assuming it based on the
    # current ASGS configuration file.
    local THIS="asgs_main.sh>checkHotstart()"
-   if [[ $FROMDIR == *"Wind10m" ]]; then 
+   if [[ $FROMDIR == *"Wind10m" ]]; then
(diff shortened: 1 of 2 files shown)


jasonfleming:
I've tested this and confirmed that it successfully resolves both issues mentioned above.

asgs_main.sh
```diff
@@ -2015,7 +2015,7 @@ if [[ $START = coldstart ]]; then
    fi
    # prepare hindcast control (fort.15) file
    # calculate periodic fux data for insertion in fort.15 if necessary
-   if [[ $PERIODICFLUX != null ]]; then
+   if [[ $PERIODICFLUX != 'null' && $FLUXCALCULATOR != 'static' ]]; then
```
oodler577, line 2018: `FLUXCALCULATOR` is first defined in `config/forcing_defaults.sh` with the value of `static`.

wwlwpd (approved):
looks good to me

Issue #1440, opened 2025-02-11
add support for `GETINPUT=some-script.sh` when fetching input files for the first time


wwlwpd:
For Mesh inputs that are stored on GH, which doesn't easily allow for large files, when ASGS detects that there is a `asgs-lfs.sh` script in the main input directory, it should prefer to run that assuming it is meant to fetch whatever files are needed; this would save the Op from having to remember to manually run this script.

If `asgs-lfs.sh` is not there, there it should fall back to the normal fallback to the ASGS CDN URL.

wwlwpd:
Once I get in there, I will assess if it'd be better or additionally useful to support ssh:// URIs. Hard to tell when using GH, which partially useful in this case, but not entirely sufficient on its own because it doesn't support large files by default, hence the need for asgs-lfs.sh

wwlwpd:
This is captured in #715 and PR #1444

PR #1444, opened 2025-02-21
Add `scp://`, `ssh://` transports; improved support alternative URLs and auto input file downloads (`INITINPUT`, e.g.).

wwlwpd:
Doing this so we can get GH to see this as something @jasonfleming needs to review.

OG PR:
  https://github.com/StormSurgeLive/asgs/pull/1088

### DIFF
--- a/asgs_main.sh
+++ b/asgs_main.sh
@@ -1806,16 +1806,24 @@ hotstartBase=fort.${LUN}
 hotstartSuffix=.nc
 hotstartPath=${LASTSUBDIR}/nowcast # only for reading from local filesystem
 hotstartURL=null
+hotstartDownloadExecutable="curl"
+hotstartDownloadRedirect="--output"
 if [[ $HOTORCOLD = hotstart ]]; then
    consoleMessage "$I Acquiring hotstart file."
    # check to see if the LASTSUBDIR is actually a URL
-   urlCheck=$(expr match "$LASTSUBDIR" 'http')
-   if [[ $urlCheck -eq 4 ]]; then
+   if [[ $LASTSUBDIR =~ "http://" || $LASTSUBDIR =~ "https://" || $LASTSUBDIR =~ "scp://" ]]; then
       # always look for fort.68.nc from a URL because only a forecast
       # will be posted to a URL, and only the hotstart file that was used
       # to start the forecast will be posted ... asgs always hotstarts from
       # a fort.68 file and always writes a fort.67 file
       hotstartURL=$LASTSUBDIR
+      # If scp is to be used, the host where the hotstart (and run.properties) files
+      # are downloaded from must support public key authentication. The URL is expected
+      # to be in the form scp://tacc_tds3//full/path/to/hotstart/file
+      if [[ $LASTSUBDIR =~ "scp://" ]]; then
+         hotstartDownloadExecutable="scp"
+         hotstartDownloadRedirect=""
+      fi
    else
       # we are reading the hotstart file from the local filesystem, determine
       # whether it is from a nowcast or hindcast
@@ -1841,7 +1849,7 @@ if [[ $HOTORCOLD = hotstart ]]; then
       debugMessage "The run directory is ${RUNDIR}."
       logMessage "Downloading run.properties file associated with hotstart file from ${hotstartURL}."
       # get cold start time from the run.properties file
-      curl $hotstartURL/run.properties > $RUNDIR/from.run.properties
+      $hotstartDownloadExecutable $hotstartURL/run.properties $hotstartDownloadRedirect $RUNDIR/from.run.properties
       logMessage "$THIS: Detecting cold start date from $RUNDIR/from.run.properties."
       COLDSTARTDATE=`sed -n 's/[ ^]*$//;s/ColdStartTime\s*:\s*//p' ${RUNDIR}/from.run.properties`
       logMessage "The cold start datetime associated with the remote hotstart file is ${COLDSTARTDATE}."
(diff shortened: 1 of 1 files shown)


jasonfleming (approved):
Some comments here: 
1. The `INITINPUT` parameter could be be renamed `GETINPUT` since that is closer to what it does.  I also think this needs to be initialized to something like `"null"` or `"none"`  to indicate the common case that there is no bespoke script to get the input files, rather than checking to see if `INITINPUT` is undefined. Also the code checks to see if the `$INITINPUT` script is executable, but does not check to see if it exists, which I think we need to do first (logging a warning if it is supposed to be there but is not found). 
2. We talked about the `MESHEXT` approach, I think this could be error prone because it adds another configuration parameter, has to be configured correctly for each mesh, and has to be implemented correctly in every local repo where we store input files, for each input file we store. It would also preclude different compressions in different repos (which may happen by accident). I'd like to just have the code assume `.xz` and then if that fails, try `.gz`, and then try no extension. Down the road we could add `.zip` or `.7z` or whatever without further complexity in the code or UX. 
I have not tested this because the initialization etc to support it is in `asgs-local-assets` repos. But I am happy to approve it if it works for our clients now. Can you add my comments in this review to a new issue that we can use to refine the changes in this PR?

wwlwpd:
> Some comments here:
> 
>     1. The `INITINPUT` parameter could be be renamed `GETINPUT` since that is closer to what it does.  I also think this needs to be initialized to something like `"null"` or `"none"`  to indicate the common case that there is no bespoke script to get the input files, rather than checking to see if `INITINPUT` is undefined. Also the code checks to see if the `$INITINPUT` script is executable, but does not check to see if it exists, which I think we need to do first (logging a warning if it is supposed to be there but is not found).
> 
>     2. We talked about the `MESHEXT` approach, I think this could be error prone because it adds another configuration parameter, has to be configured correctly for each mesh, and has to be implemented correctly in every local repo where we store input files, for each input file we store. It would also preclude different compressions in different repos (which may happen by accident). I'd like to just have the code assume `.xz` and then if that fails, try `.gz`, and then try no extension. Down the road we could add `.zip` or `.7z` or whatever without further complexity in the code or UX.
>        I have not tested this because the initialization etc to support it is in `asgs-local-assets` repos. But I am happy to approve it if it works for our clients now. Can you add my comments in this review to a new issue that we can use to refine the changes in this PR?

Oh, thank you. I forgot to do that! I will update the variable name and move the `MESHEXT` to another branch and issue.

wwlwpd:
https://github.com/StormSurgeLive/asgs/issues/1446

wwlwpd:
> The INITINPUT parameter could be be renamed GETINPUT since that is closer to what it does. I also think this needs to be initialized to something like "null" or "none" to indicate the common case that there is no bespoke script to get the input files, rather than checking to see if INITINPUT is undefined. Also the code checks to see if the $INITINPUT script is executable, but does not check to see if it exists, which I think we need to do first (logging a warning if it is supposed to be there but is not found).

I changed the variable to `GETINPUT`, and if not set defaults to `null`; the dispatch code also checks to see if it `-e` and `-x` seperately, warns if it doesn't exist and if it exists but is not executable.

The execution protocol remains the same, i.e.:

# `cd $INPUTDIR`
# `./$GETINPUT`

This is to simplify the script, it just downloads the files in the present working directory (`$INPUTDIR`.)

Please let me know what you think.

wwlwpd:
With, `GETINPUT=get-lfs.sh` (e.g.):

image

then subsequently (with `get-lfs.sh` I am using as the test):

image

Then with it `GETINPUT=null`, using the `MESHURL` and similar,

image

wwlwpd:
Test with this local assets config, https://github.com/CERA-GROUP/asgs-local-assets/blob/issue-1/config/mesh_defaults.sh

image

jasonfleming:
Thank you @wwlwpd !! I will test :-)

wwlwpd:
@jasonfleming just wanted to ping you about this one

jasonfleming:
FYI, I ran into the following during the build:
```
Cloning into '/home/jason/Campaigns/Development/2025/issue-715/git/asgs-lint'...
Warning: the ECDSA host key for 'github.com' differs from the key for the IP address '140.82.113.3'
Offending key for IP in /home/jason/.ssh/known_hosts:28
Matching host key in /home/jason/.ssh/known_hosts:46
Are you sure you want to continue connecting (yes/no)? 
```
Could be an issue for batch/unattended installations?

wwlwpd:
> FYI, I ran into the following during the build:
> 
> ```
> Cloning into '/home/jason/Campaigns/Development/2025/issue-715/git/asgs-lint'...
> Warning: the ECDSA host key for 'github.com' differs from the key for the IP address '140.82.113.3'
> Offending key for IP in /home/jason/.ssh/known_hosts:28
> Matching host key in /home/jason/.ssh/known_hosts:46
> Are you sure you want to continue connecting (yes/no)? 
> ```
> 
> Could be an issue for batch/unattended installations?

I get that on qbc - though I have not tracked it down, I can say it's an issue between your environment and Github; it's not related directly to any repository.

jasonfleming:
Yeah I got this on my local desktop machine. Hopefully they change host keys on their servers rarely enough that we won't have builds stopping and waiting for a manually typed  "yes" very often. :-)

jasonfleming:
The host key thing was not a show stopper, I deleted the offending key on my machine and it was all good.

wwlwpd:
> Yeah I got this on my local desktop machine. Hopefully they change host keys on their servers rarely enough that we won't have builds stopping and waiting for a manually typed "yes" very often. :-)

I agree, it's annoying and I've been putting up with it on qbc. I need to see what the root cause is, it's upstream but still would be nice to know a quick solution when it happens (or how to avoid it.)

wwlwpd:
@jasonfleming I rebased this with `master` after today's new merges.

wwlwpd:
@jasonfleming - this PR has just the changes we were proposing for #715, with the additional support for `GETINPUT` support for "lfs" (#1440).

jasonfleming:
@wwlwpd I tested this with a fresh install by hotstarting from a remote hotstart file. I was able to do this successfully with `https` but my attempt to use `ssh` failed. I took a look at the relevant section of the code in `asgs_main.sh` and found that the changes that implement this behavior for hotstart files in commit `e1b3def` were not present for some reason. How can we add this back?

wwlwpd:
@jasonfleming I added the first 2 commits, I am not sure how the got reset but they are in the history. I used `cherry-pick` - can you see if there are any conflics?

image

jasonfleming (approved):
Ok I have tested this and found that it resolves #715. Thank you!

PR #1462, opened 2025-03-25
Loosened check on comparison of RunStartTime and RunEndTIme

### LINKED ISSUE #1377: Use existing nowcast for tropical cyclones when a new one is not necessary
When constructing a nowcast for a new forecast/advisory, the actual hotstart file may already be up to date, depending on when the previous BEST track file was issued relative to when the previous forecast/advisory ran. If this is the case, then the hotstart file from the previous nowcast should be used for the new forecast jobs.

jasonfleming:
resolves #1377

### DIFF
--- a/asgs_main.sh
+++ b/asgs_main.sh
@@ -2638,7 +2638,7 @@ while [ true ]; do
    s=${properties['RunStartTime']}
    e=${properties['RunEndTime']}
    # Try to calculate lengthCheck; default to 0 on error
-   if ! lengthCheck=$(( e - s )) ; then
+   if ! lengthCheck=$(( e - s )) ; then # fails if $e or $s is not an int or unset
       # can't recover from either of the values of these parameters being anything other than 10 digit integers
       fatal "Could not compute nowcast run length. In '$PWD/${SCENARIO}.run-control.properties' the RunStartTime is'$s' and RunEndTime is '$e'."
    fi
--- a/asgs_main.sh
+++ b/asgs_main.sh
@@ -2638,7 +2638,7 @@ while [ true ]; do
    s=${properties['RunStartTime']}
    e=${properties['RunEndTime']}
    # Try to calculate lengthCheck; default to 0 on error
-   if ! lengthCheck=$(( e - s )) 2>/dev/null; then
+   if ! lengthCheck=$(( e - s )) ; then
       # can't recover from either of the values of these parameters being anything other than 10 digit integers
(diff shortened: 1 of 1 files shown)


wwlwpd (changes requested):
@jasonfleming can we make the code more robust? I've noticed using `-gt` is fragile, can we consider something like the following? My concern is that `$lengthCheck` can be a non-number (or empty), and this has caused other issues in the past. For example,

```
start=${properties['RunStartTime']}
end=${properties['RunEndTime']}

# Try to calculate lengthCheck; default to 0 on error
if ! lengthCheck=$(( end - start )) 2>/dev/null; then
  lengthCheck=0 # or handle invalid input here ...
fi

# Main logic
if [[ $lengthCheck -gt 1 ]]; then
  logMessage "$ENSTORM: $THIS: Starting nowcast for cycle '$ADVISORY'."
  ...
```

jasonfleming:
OK I have made changes as described, and tested successfully. However, in manual (command line) tests of that specific code snippet, I had a hard time getting the conditional to return a `lengthCheck` value of zero. The arithmetic operation always seemed to succeed, even if `start` and `end` were set to random letters, etc.

wwlwpd:
@jasonfleming I reproduced it (`lengthCheck` is getting set to `0` in this case, which is correct;) and did some more looking, this should work as intended, in fact you should be able to remove the capture of the STERR. I will leave a change that suggestion that you can take or leave.

asgs_main.sh
```diff
@@ -2635,7 +2635,14 @@ while [ true ]; do
    # load properties
    declare -A properties
    loadProperties ${SCENARIO}.run-control.properties
-   if [[ ${properties['RunEndTime']} != ${properties['RunStartTime']} ]]; then
+   s=${properties['RunStartTime']}
+   e=${properties['RunEndTime']}
+   # Try to calculate lengthCheck; default to 0 on error
+   if ! lengthCheck=$(( e - s )) 2>/dev/null; then
```
wwlwpd, line 2641:
```suggestion
   if ! lengthCheck=$(( e - s )); then # fails if $e or $s or unset or not an integer
```

wwlwpd (approved):
This change looks good; I recommend you remove the redirection of `STDERR` since it is not needed. But I am approving anyway.

jasonfleming:
OK I have removed the redirect to stderr as described. Looks like the previous review/approval was automatically dismissed when I made changes? Please review and merge at your convenience if approved.

wwlwpd (approved):
looks good! ty!

Issue #1333, opened 2024-06-08
SWAN is being built with ifort on all the LSU machines except QBD, where it is being built with gfortran


wwlwpd:

jasonfleming:
Need to confirm whether this is actually an issue. But should be resolved if it actually is.

jasonfleming:
I just checked this in the latest release candidate `2025.1.rc-6` and found that it is resolved successfully.

PR #1468, opened 2025-04-08
Ensure swan's macros.inc is updated properly for ADCIRC v53

wwlwpd:
@jasonfleming - while investigating things further on stampede3, I created a patch that I think ensures that swan/macros.inc (for v53) is updated properly. I am creating a PR for us to look over. This didn't fix the issue I was seeing on stampede3, but this is the kind of deep dive I think we were wanting to do with #1333. I need to go through and apply patch updates the other supported ADCIRC versions once we get this worked out. Note: the other versions move SWAN stuff to the `thirdaparty/swan` directory.

Related to #1333.

### DIFF
--- a/bin/init-adcirc.sh
+++ b/bin/init-adcirc.sh
@@ -32,6 +32,29 @@ if [ "${1}" = "clean" ]; then
   exit 0
 fi
 
+# get compiler info
+case "$ADCIRC_COMPILER" in
+intel)
+  # default
+  _FC=fort
+  _CC=icc
+;;
+intel-oneapi)
+  # default, above
+  _FC=ifort
+  _CC=icx
+;;
+gfortran)
+  _FC=gfortran
+  _CC=gcc
+;;
+*)
+  echo '${W} unknown compiler is unsupported...; defaulting to "intel"'
+esac
+_FC_VERSION=$($_FC --version | head -n 1)
+_CC_VERSION=$($_CC --version | head -n 1)
+
+
 while getopts "bN:" optname; do
    case $optname in
       b) BATCH=1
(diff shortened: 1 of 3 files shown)


wwlwpd:
@jasonfleming - as nice as it is to have this fully consistent, it did not help with SAB on stampede3. However, I was able to run EGOM just fine with v53. So I think it has something to do with the input file support for SAB "default" for v53.05 and v55.02 on stampede3 specifically.

Let me know how you'd like to proceed - looking at v56.0.2's `thirdparty/swan/macros.inc` file was certainly interesting. The `thirdparty/swan/platform.pl` has some notable updates in it compared to what is in v53.05. I am interested in getting some direction about what we should do here since after all of this, I am not convinced that the way v53.05 is being built on stampede3 has much to do with the errors.

E.g., we could go with these updates for v53.05 as a first pass and make similar adjustments to v55.02 and v56.0.2 (or v56.0.3?), but then look at applying the updates in `macros.inc` across the board. In all cases, I do thing SWAN's `platform.pl` is poorly written, but as it stands could use some [ninja-code](https://www.perlmonks.org/?node_id=388433) updating without giving into the tempation to "update" the whole thing. Anyway, let's discuss since I do not think these changes are critical, but seem desirable. We also must consider our plans to update to v56.0.3 soon.

jasonfleming:
@wwlwpd My preference at this point is to have our build automation pick the right `macros.inc` rather than rely on the stock scripts that come with SWAN. Not sure if that makes things easier or more difficult. Let's talk about this in our meeting tomorrow.

patchtrends:
> @wwlwpd My preference at this point is to have our build automation pick the right `macros.inc` rather than rely on the stock scripts that come with SWAN. Not sure if that makes things easier or more difficult. Let's talk about this in our meeting tomorrow.

This is fine wiht me, however note that the scripts are called by both ADCIRC's `work/makefile` and SWAN's `Makefile` - `platform.pl` and `switch.pl`, respectively. So we're either going to do our scripts be aware of this or take the calls out of the respective makefiles and call them ourselves - at least `platform.pl`. We will also either need to modify `platform.pl` to handle additional platforms as we move forward in the future or delete `platform.pl` altogether and do our own complete `macros.inc` management. I have no preference, just giving a lay of the land.

perlsheperd:
This is going to be superceded by #1471.

PR #1474, opened 2025-04-12
Refactoring parameter packages

### LINKED ISSUE #1472: Create `parameterPackage` subvariant capability
There are some related changes to support more flexible parameter packages for different modeling scenarios. These are as follows:

1. Create a `subvariant` capability so that a `parameterPackage` can have one or more subvariants without making a whole new parameter package. For example, a parameter package can be specified as usual and have a subvariant within it that only changes the lateral turbulence formulation (e.g., eddy viscosity in the main package vs that is changed to smagorinsky in the subvariant with no other changes).
2. Use `2025.1` as the `parameterPackage` that is typically used for a mesh because `default` seems to indicate that it would be used if nothing at all was specified (that behavior would not backward compatible with older asgs config files). The use of `default` would still work the same as it does now but its use would be deprecated.
3. Record the value of `parameterPackage` and `subvariant` (if any) in metadata for monitoring and troubleshooting.

jasonfleming:
Reduced code in `config/mesh_defaults.sh` to `source` parameters for each mesh from dedicated scripts (one for each mesh); created a parameter package file for NGOMv19b that contains a `subvariant` set of parameters  for testing physics as introduced in #1466 

resolves #1472


wwlwpd:
Looks like a great start, @jasonfleming - did you not add the other files yet?

wwlwpd:
If you want to do one mesh so I have an example, I could also help with doing the rest. Regarding `$ASGS_LOCAL_DIR` (local assets), as long as `config/mesh_defaults.sh` remains and supports a custom `$ASGS_LOCAL_DIR/config/mesh_defaults.sh`, there should be no need to do anything else to support local assets with meshes organized in this way. We should not worry about someone with local assets organizes their meshes. If they want add support for `parameterPackages` and `subvariant`s, then that is one them. Make sense?

jasonfleming:
Ok @wwlwpd that sounds great. We already have `$SCRIPTDIR/input/meshes/*` directories that contain `fort.15` templates etc for each mesh ... should that info be combined into the `$SCRIPTDIR/meshes` directory, perhaps in a future PR?

wwlwpd:
> Ok @wwlwpd that sounds great. We already have `$SCRIPTDIR/input/meshes/*` directories that contain `fort.15` templates etc for each mesh ... should that info be combined into the `$SCRIPTDIR/meshes` directory, perhaps in a future PR?

Actually, I would scratch what I said about `$SCRIPTDIR/meshes` and use the already existing mesh directories in `$SCRIPT/input/meshes`. So for each mesh directory, have a `parameterPackages.sh` and an `init.sh` (or you could put it all in `init.sh`, your call). The `about.txt` files are something I can add later when I need it. It's not needed here.

jasonfleming:
Hey @wwlwpd I added the `init.sh` and `about.txt` files to the existing mesh directories under `$SCRIPTDIR/input/meshes` as suggested. The parameter packages are included in the `init.sh` files. The `about.txt` files are just placeholders, containing only the standard name of the mesh. Please let me know what you think.

wwlwpd (approved):
Looks good to me!

jasonfleming:
OK once I test this, I will take it out of draft and it will be ready to merge.

jasonfleming:
This is running as expected with NGOMv19b. Will test some other meshes also.

wwlwpd (approved):
Great change, thank you.

PR #1491, opened 2025-05-06
Updating NCEP URLs

wwlwpd:
Issue 1479:

* get_nam.pl doesn't appear to need any updates, it is already using the correct URL and path for the "nomads" HTTPS server
* bin/get_gfs.pl was using one of the HTTPS interfaces for the ftp server, so this was updated to use the correct "nomads" server

WIP - need test and vet any other scripts this may affected.

### DIFF
--- a/bin/get_gfs_status.pl
+++ b/bin/get_gfs_status.pl
@@ -57,7 +57,6 @@ GetOptions(
     # replaces the $ftp->ls for files, looks only for gfs.* files
     sub http_ls {
         my $dir = shift;
-        $dir =~ s/\/pub\///g;
         my $url         = sprintf( qq{https://nomads.ncep.noaa.gov/%s/ls-l}, $dir );
         my $res         = $ua->get($url);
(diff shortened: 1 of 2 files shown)


wwlwpd:
I started a GFS forced SAB run like the NAM that is currently running on QBD, since it appears as if the only necessary changes are going to be for GFS.

wwlwpd:
not quite ready

wwlwpd:
looks ok

wwlwpd:
QBD shutting down unexpectedly around the time of this switch over upstream has made it more difficult to ascertain the change/breakage, but I am monitoring the runs there to see.

wwlwpd:
Currently running in daily mode for GFS and NAM. Ready for review, but should also be tested through another daily run or so.

wwlwpd:
This PR appears to be working for me. It's probably safe to merge since the alternative if meteorology not working at all.

PR #1495, opened 2025-05-20
Made bin/fetch aware of origin's remote protocol (https or ssh)

### LINKED ISSUE #1494: `fetch` should detect remote protocol type
i.e., it should detect when ASGS was checked out using `https` or `git@` (`ssh`), and adjust the base url used accordingly.

wwlwpd:
Issue 1494: added some flexibility to fetch, for when the ASGS local directory from an https based URL or .zip, fixed a bug with 'fetch cera'

Resolves #1494

### DIFF
--- a/bin/fetch
+++ b/bin/fetch
@@ -9,8 +9,23 @@ if [ -z "$_ASGSH_PID" ]; then
   fi
 fi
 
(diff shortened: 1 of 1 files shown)


wwlwpd:
Generally applicable enhancement that fell out of the adcirclive branch.

PR #1504, opened 2025-06-04
Updated handler for `list meshes`

### LINKED ISSUE #1503: update `list meshes` to show nodes for new mesh directory hierarchy
based on,

```
find . -name init.sh | xargs grep nodes | awk -F= '{print $2 " " $1}' | sort -nr
```

Image

wwlwpd:
Issue 1503: the directory structure for storing configurations for officially supported meshes changed, but this impacted the `list meshes` command in a way that stopped the number of nodes from being reported. This PR fixes that.

Meshes that are contained in local assests, if set, are handled as they were before using the original method.

Resolves #1503.

### DIFF
--- a/bin/list
+++ b/bin/list
@@ -36,10 +36,9 @@ _list() {
       fi
       ;;
     meshes)
-      for m in $(cat $ASGS_MESH_DEFAULTS | grep '")' | sed 's/[")]//g' | awk '{print $1}'); do
-        printf "% 2d. %s\n" $LISTNUM $m
-        LISTNUM=$(($LISTNUM+1))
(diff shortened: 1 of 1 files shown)


wwlwpd:
Minor issues, but useful when looking for meshes of a particular size. To test:

1. `git fetch --all`
2. `git checkout issue-1503`
3. `./asgsh`
4. issue command, `list meshes`
5. should look like the picture in the original issue, #1503

jasonfleming:
Thank you, testing this now.

jasonfleming:
It works!

PR #1506, opened 2025-06-09
Moved quality checks out of queue script

### LINKED ISSUE #1473: Relocate solution quality checks from queue script to hook script
The implementation of solution quality checks in queue scripts may fail due to dependencies that do not resolve as expected on compute nodes. The solution quality checking should be implemented as hook scripts instead, since these execute on non-compute nodes and will have their dependencies resolved.

jasonfleming:
resolves #1473

### DIFF
--- a/asgs_main.sh
+++ b/asgs_main.sh
@@ -2745,6 +2745,7 @@ while [ true ]; do
       THIS="asgs_main.sh"
       # check once per minute until all jobs have finished
       monitorJobs "$QUEUESYS" "${JOBTYPE}" "${ENSTORM}" "$NOWCASTWALLTIME"
+      $SCRIPTDIR/util/output/quality.sh $JOBTYPE $SCENARIO $(<jobID) $CYCLE $SCENARIO $SCRIPTDIR >> $SCENARIOLOG # padcirc nowcast 555 2024072406 nowcast $SCRIPTDIR
       THIS="asgs_main.sh"
       # check to see that the nowcast job did not conspicuously fail
       handleFailedJob $RUNDIR $ADVISDIR $ENSTORM ${OUTPUTDIR}/${NOTIFY_SCRIPT} $HPCENV $STORMNAME $YEAR $STORMDIR $ADVISORY $LASTADVISORYNUM $STATEFILE $GRIDFILE $EMAILNOTIFY "${JOB_FAILED_LIST}" $ARCHIVEBASE $ARCHIVEDIR
@@ -3230,6 +3231,8 @@ while [ true ]; do
       (
          monitorJobs "$QUEUESYS" "${JOBTYPE}" "${ENSTORM}" "$FORECASTWALLTIME"
          THIS="asgs_main.sh"
+         # check quality of results
+         $SCRIPTDIR/util/output/quality.sh $JOBTYPE $SCENARIO $(<jobID) $CYCLE $SCENARIO $SCRIPTDIR >> $SCENARIOLOG # padcirc nowcast 555 2024072406 nowcast $SCRIPTDIR
          handleFailedJob $RUNDIR $ADVISDIR $ENSTORM ${OUTPUTDIR}/${NOTIFY_SCRIPT} $HPCENV $STORMNAME $YEAR $STORMDIR $ADVISORY $LASTADVISORYNUM $STATEFILE $GRIDFILE $EMAILNOTIFY "${JOB_FAILED_LIST}" $ARCHIVEBASE $ARCHIVEDIR
          THIS="asgs_main.sh"
          # only attempt post processing if this scenario
--- a/qscript.template
+++ b/qscript.template
@@ -85,7 +85,6 @@ fi
 #                 L O O P   O V E R   L A Y E R S
 #----------------------------------------------------------------------------
 #
-
 for layer in ${layers[@]}; do
    if [[ -e "fort.15" ]]; then
       rm fort.15 2>> $SCENARIOLOG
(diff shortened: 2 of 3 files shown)


asgs_main.sh
```diff
@@ -2745,6 +2745,7 @@ while [ true ]; do
       THIS="asgs_main.sh"
       # check once per minute until all jobs have finished
       monitorJobs "$QUEUESYS" "${JOBTYPE}" "${ENSTORM}" "$NOWCASTWALLTIME"
+      $SCRIPTDIR/util/output/quality.sh $JOBTYPE $SCENARIO $(<jobID) $CYCLE $SCENARIO $SCRIPTDIR >> $SCENARIOLOG # padcirc nowcast 555 2024072406 nowcast $SCRIPTDIR
```
wwlwpd, line 2748: I know this is currently a draft, but did we want to add this as POSTPROCESS hook or create a new "QAQC" hook?

jasonfleming: Great question. That was my first inclination also. However, I took a look at how we're currently setting hook scripts, with the Operator config file specifying them as a bash array with `=` rather than `+=`, e.g.:
```bash
POSTPROCESS=( includeWind10m.sh createOPeNDAPFileList.sh $OPENDAPPOST ) 
hooksScripts[FINISH_SPINUP_SCENARIO]=" output/createOPeNDAPFileList.sh output/$OPENDAPPOST " 
hooksScripts[FINISH_NOWCAST_SCENARIO]=" output/includeWind10m.sh output/createOPeNDAPFileList.sh output/$OPENDAPPOST " 
```
A new QAQC hook could work, but it seemed redundant with `FINISH_SPINUP_SCENARIO`, `FINISH_NOWCAST_SCENARIO`, and `POSTPROCESS`.

I didn't want the Operators to have to specify `quality.sh` to avoid the additional burden/complexity when creating a config file.

I also didn't want the `quality.sh` script to be left out (accidentally or intentionally).

I couldn't think of a way to have `quality.sh` added by default in the right places and still enable Operators to specify their own hook scripts in the Operator's config file.

Advice on improving this approach is welcome.

perlsheperd:
@jasonfleming ping me when ready for a review

wwlwpd:
@jasonfleming ping me when ready for a review

asgs_main.sh
```diff
@@ -1312,6 +1312,18 @@ monitorJobs()
       postScenarioStatus
    fi
    sleep 30 # give buffers a chance to flush to the filesystem
+   if [[ $JOBTYPE == "padcirc" || $JOBTYPE == "padcswan" ]]; then
+      if [[ $QUALITYCONTROL != "off" ]]; then
```
wwlwpd, line 1316:
```suggestion
      if [[ "$QUALITYCONTROL" != "off" ]]; then
```

asgs_main.sh
```diff
@@ -1312,6 +1312,18 @@ monitorJobs()
...
+         logMessage "$ENSTORM_TEMP: $THIS: Running quality control checks."
+         if [[ ! -e $QUALITYCONTROL ]]; then
```
wwlwpd, line 1318:
```suggestion
         if [[ ! -x "$QUALITYCONTROL" ]]; then
```

asgs_main.sh
```diff
@@ -1312,6 +1312,18 @@ monitorJobs()
...
+            warn "$ENSTORM_TEMP: $THIS: The quality control script '$QUALITYCONTROL', specified by the QUALITYCONTROL parameter, was not found."
```
wwlwpd, line 1319:
```suggestion
            warn "$ENSTORM_TEMP: $THIS: The quality control script '$QUALITYCONTROL', specified by the QUALITYCONTROL parameter, was not found or is not executable."
```

asgs_main.sh
```diff
@@ -1312,6 +1312,18 @@ monitorJobs()
...
+         else
+            $QUALITYCONTROL >> $SCENARIOLOG
+         fi
+      else
+         logMessage "$ENSTORM_TEMP: $THIS: Quality control checks have been deactivated by setting QUALITYCONTROL='off'."
```
wwlwpd, line 1324:
```suggestion
         logMessage "$ENSTORM_TEMP: $THIS: Quality control checks have been deactivated by setting: QUALITYCONTROL=off"
```

wwlwpd (changes requested):
Please take a look at the minor tweaks I've suggestion. Overall, looks great! I am looking forward to using the hook functions.

asgs_main.sh
wwlwpd: this change goes with the one right before it (`-x`)

wwlwpd:
> ty @wwlwpd I will make those changes ... what about the functions to set the hook scripts in `manageHooks.sh` ... right now I have a fully implemented generic pair (to add and to remove) and a bunch of function stubs for the hooks that I did not implement b/c it seemed like a lot of repeated code and I wanted to get your opinion.

Ah, well I am okay with how you have it. It'd be nice to go in and fill then in; but having the general basis as a common would be good, for example:

```
removeScriptFrom()
{
    local hook=$1
    local hookScript=$2
    local s=""          # list of scripts for this hook
    local THIS="asgs_main->manageHooks->removeScriptFrom_FINISH_NOWCAST_SCENARIO()"
    for k in ${hookScripts["${hook}"]}; do
        if [[ $k != "$hookScript" ]]; then
            s+=" $k"
        fi
    done
    hookScripts["${hook}"]="$s"
}

removeScriptFrom_FINISH_NOWCAST_SCENARIO()
{
    local hookScript=$1
    removeScriptFrom "FINISH_NOWCAST_SCENARIO", "$hookScript"
}
```

But, I do think that it'd be good to have explicit add/remove functions for each hook, for clarity down the line.

asgs_main.sh
jasonfleming: Fixed now thank you!

asgs_main.sh
jasonfleming: This is now implemented in a way that allow us to turn it off.

PR #1509, opened 2025-06-13
Updating emails to include more run-specific information.

### LINKED ISSUE #1502: add `<MESH>_<FORCING>_<MACHINE>_<OPERATORID>` to `opendap_post2.sh` email subject for easier identification
Instead of updating the default `INSTANCENAME` add this information to the email subject that get sent from `opendap_post2.sh`.

wwlwpd:
Issue 1502: Uses now,

   <MESH>_<FORCING>_<MACHINE>_<OPERATORID>

Instead of:

   <MACHINE>.<ASGSADMIN>

Deleted opendap_post.sh, which has been long deprecated.

Resolves #1502.

### DIFF
--- a/input/meshes/TXLA22a/TXLA22a.15.ASGS2024.1.template
+++ b/input/meshes/TXLA22a/TXLA22a.15.ASGS2024.1.template
@@ -1269,5 +1269,4 @@ K2
 !
 ! namelists:
 %met_control_namelist%
-%wetdry_control_namelist%
 %inundation_output_control_namelist%


wwlwpd:
@wwlwpd it turn that unless we set `ASGSADMIN_ID` in the config (or `run.properties`), then it'll use `ASGSADMIN`:

image

This was already accounted for in `opendap_post2.sh`.

But I think we got it. I had to detect and add `gahm` where `nam` or `gfs` were added otherwise, unless this is already accounted for in `run.properties` via `${properties["forcing.nwp.model"]}`?

PR #1516, opened 2025-06-25
Explicitly disable GSL for NCO.

### LINKED ISSUE #1515: disable GSL for NCO in `./configure` step

wwlwpd:
Issue 1515: Adds --enable-gsl=no to the ./configure step, thus unshackling us from the tyranny of an unnecessary dependency.

Resolves #1515.


jasonfleming:
Thank you, my understanding is that the only reason `nco` depends on `libgsl` is to support `ncap2`, which I am in the process of eliminating the need for. Thank you for implementing this fix to avoid this unnecessary dependency.

PR #1530, opened 2025-07-08
Updating patches for "debian+slurm".

### LINKED ISSUE #1529: update ADCIRC patches to reflect `debian+slurm` ADCIRC cluster

wwlwpd:
Issue 1529:

Resolves #1529

### DIFF
--- a/patches/ADCIRC/v56.0.2/01-cmplrflags-mk.patch
+++ b/patches/ADCIRC/v56.0.2/01-cmplrflags-mk.patch
@@ -1,5 +1,5 @@
 diff --git a/work/cmplrflags.mk b/work/cmplrflags.mk
-index 9f89de7..3435029 100644
+index 9f89de78..4d3d7932 100644
 --- a/work/cmplrflags.mk
 +++ b/work/cmplrflags.mk
 @@ -1,5 +1,46 @@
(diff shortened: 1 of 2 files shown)


jasonfleming:
I have verified that all supported ADCIRC versions build successfully with intel-oneapi. I am now trying again with gfortran just for completeness.

wwlwpd:
> Not sure whether this should be in the same PR or a new one but here are some comments:
> 
>     1. The `HPCENVSHORT` is set to `debian` in `platforms/debian+slurm/init.sh` instead of `debian+slurm` so these two platforms will have the same `$HPCENVSHORT`
> 
>     2. The default compiler for `debian+slurm` is set to `gfortran` ... should it be `intel-oneapi`?
> 
>     3. The batch job ID is not being captured correctly when a job is submitted, which scrambles the job statuses in `scenario.status.json` ... the `debian+slurm` platform needs to be added to `monitoring/captureJobID.sh` in the same section as stampede3 etc
> 
>     4. I am building and installing all ADCIRC versions with the `intel-oneapi` compiler ... would like to try them all with `gfortran` as well

Thank you, I keep getting distracted from addressing this. I also built all the oneAPI versions on QBD successfully:

1. I'll update that
2. I think we should keep it gfortran so users don't have to install oneAPI
3. I will update that, ty!
4. thank you!

jasonfleming:
One other thing is that the cluster where I am testing is not automatically detected as `debian+slurm` ... `init-asgs.sh` thinks it is just `debian`. 

Finally, I have successfully built all supported versions of ADCIRC with both `intel-oneapi` and `gfortran` so I am approving this PR. :-)

jasonfleming (approved):
It works!! Please merge. :-)

wwlwpd:
@jasonfleming - please re-review/approve, I address 1, 3, and `bin/guess` as pointed out above - thanks for catching all these! I don't think it's necessary to rebuild or try to build ADCIRC again (you can if you want obviously). Thanks!

PR #1556, opened 2025-08-29
Adding active nodal attributes and their default values to `run.properties`

### LINKED ISSUE #1550: Add active nodal attributes and their default values to `run.properties`

jasonfleming:
resolves #1550, testing now


jasonfleming:
This PR is not needed after all.

PR #1567, opened 2025-09-09
Refined/relaxed quality control checks

### LINKED ISSUE #1566: Refine quality checking for nonfatal numerical instability and allowance of min value in maxele to be exactly zero
Quality checking sometimes reveals corner cases such as model results that are still usable despite localized nonfatal numerical instabilities as well as the minimum value in a maxele file being exactly zero without other issues. The quality check should be refined so that the Operator can configure it to accept nonfatal numerical instability messages. The checks on water surface elevation files should be relaxed to allow the minimum value in the file to be exactly zero.

jasonfleming:
resolves #1566


jasonfleming:
still need to test

wwlwpd (approved):
Looks good!

Issue #567, opened 2021-07-15
ASGS runstart time out of sync with advisory time


carolakaiser:
During storms, when moving an ASGS instance from one HPC to another, it often happens that the initial first run on the new machine has a delay. This delay causes the runstart time for the ASGS run to be later than the actual advisory time. E.g. when the advisory time is 09 AM, the runstart time is normally 06 AM, however in such cases it is 12 AM or later. While this is not necessarily an error, it causes problems for CERA, especially when there are already runs posted that have the 06 AM time as runstart but the same advisory time.
CERA has a limitation with handling such runs properly and with that a lot of manual clean-up is necessary. Is there a way to test in ASGS if the runstart time is set to a later time than the advisory time? And if so, can these runs maybe marked in some way (like with a value in the run.properties file) so that those runs could be rejected by CERA?

carolakaiser:
Now, when thinking more about it, I could of course do such test in CERA on my own :-). Okay, so the better question is, would you see this behavior in ASGS as a correct one and does that need to be addressed in CERA?

wwlwpd:
Can you recommend a field name and range of values?

One thought also is that we need to conduct an audit of how we handle time zones (e.g., CDT vs GMT/UTC)

carolakaiser:
Since I do not exactly know whether this is an ASGS issue or not, I have decided to implement the test in CERA. If you think that ASGS works correctly, please close this ticket.
As far as time zones, I don't think that ASGS needs to deal with that? In fact, I would recommend keeping everything in UTC to stay consistent.

wwlwpd:
Thanks, just thinking that we need to do an audit to make sure.

wwlwpd:
@jasonfleming - can you add the information you provided at the meeting today? Thanks.

jasonfleming:
@wwlwpd my initial impression based on Carola's description is that we hotstarted from a URL that had a hotstart file that was later than the advisory time. In these cases, the ASGS will ignore the forecast information that is earlier than the hotstart file, and start the forecast run at the hotstart time, with the forecast data that are applicable to the hotstart time. So, if the forecast was issued at 03Z on July 1st, and we start ASGS with a hotstart file for 03Z on July 2nd, ASGS will set up the ADCIRC run using the forecast data that starts on July 2nd (and thereafter). Side effects include only having numerical results on July 2nd and thereafter and the forecast run being 1 day shorter than the given forecast duration.

wwlwpd:
[...]

PR #1583, opened 2025-10-09
Support for RRFS

jasonfleming:
This is a WIP, most of the nowcast code is complete and tested successfully. Still need to finish the nowcast scripting and then add/test/deploy some code to cover forecasts.


jasonfleming:
most recent commit resolves #567

wwlwpd (approved):
Looks good! Thanks for all the work here. Adding `jq` to the initial build was a good call.

PR #1586, opened 2025-10-19
Updating HDF5 and netCDF4 to more recent versions

wwlwpd:
Issue 1585: artifact of recent work troubleshooting issues present on QBD, SuperMikeIII, and Stampede.

### DIFF
--- a/cloud/general/init-hdf5.sh
+++ b/cloud/general/init-hdf5.sh
@@ -52,9 +52,8 @@ chmod 700 $_ASGS_TMP
 cd $_ASGS_TMP
 
 if [ ! -e hdf5-${HDF5_VERSION}.tar.gz ]; then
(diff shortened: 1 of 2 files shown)


wwlwpd:
This was done to see if getting HDF5 and netCDF4 in line with the Singularity version that works on QBD (which is actually CentOS 8 based, not RHEL) solved the issue. It does not, but this might be a good time to get in line with the latest upstream versions and to use the upstream sources, unless we have reason to continue to use our own.

jasonfleming:
Hey @wwlwpd it should not affect the numerical solutions at all. My suggestion would be to build/install this branch on one of our HPC platforms and run a full size mesh (e.g., HSOFS) with GFS for a few cycles. If there are no surprises in any of that, I think it should be good to go.

wwlwpd:
This seems to be working fine, @jasonfleming - https://fortytwo.cct.lsu.edu/thredds/fileServer/netcdfUpgradeTest/2025/GFS/

Issue #1588, opened 2025-10-22
Support registration of externally built ADCIRC binaries, must be Singularity aware (queue script gen/submit, etc)


wwlwpd:
"Singularity Aware, means:

- ADCIRC binaries inside of a singularity container require special invocation via Slurm's `srun` on SLURM
- add an environmental variable to explicitly denote if it is singular or not

Singularity also has some assumed restrictions:

1. singularity is not available on head nodes
2. singularity is available on compute nodes
3. `verify adcirc` will not work because this is run on the local computing environment
4. initially support just Slurm
5. when a singularity based ADCIRC is "loaded" (via `load adcirc`), no ADCIRC utilities are going to be available locally via `PATH` since singularity won't run on a local head node (at LSU, but assuming everywhere)

PR #1601, opened 2025-11-15
Issue #1588 - initial Singularity support

wwlwpd:
intentionally blank

### DIFF
--- a/bin/init-adcirc.sh
+++ b/bin/init-adcirc.sh
@@ -306,6 +306,21 @@ if [ "$BUILD_PARALLEL_ADCIRC" != "yes" ]; then
   BUILD_PARALLEL_ADCIRC=no
   __ADCIRC_PROFILE_NAME=${__ADCIRC_PROFILE_NAME}-DEBUG-${DEBUG}-serial-only
 fi
+echo
+
+#
+# S I N G U L A R I T Y  S U P P O R T
+#
+
+__ADCIRC_SINGULARITY_SIF=
+if [[ -n "$ASGS_SINGULARITY_CMD" && -z "${BATCH}" ]]; then
+  read -p "Enter full path to the .sif (<enter> if you don't know)? " _ADCIRC_SINGULARITY_SIF
+fi
+
+if [ -n "${_ADCIRC_SINGULARITY_SIF}" ]; then
+  ADCIRC_SINGULARITY_SIF=$_ADCIRC_SINGULARITY_SIF
+  __ADCIRC_PROFILE_NAME=${__ADCIRC_PROFILE_NAME}-singularity
+fi
 
 #
 # C H O O S E  A D C I R C  P R O F I L E  N A M E
@@ -582,7 +597,9 @@ $patchJSON
     "adcirc.build.c.compiler.version"     : "$_CC_VERSION",
     "adcirc.build.modules.loaded"         : "$MODULE_LIST",
     "adcirc.build.fortran.compiler.version" : "$_FC_VERSION",
-    "adcirc.build.debug"                  : "$DEBUG"
+    "adcirc.build.debug"                  : "$DEBUG",
+    "adcirc.singularity.sif"              : "$ADCIRC_SINGULARITY_SIF",
+    "asgs.container.cmd"                  : "$ASGS_SINGULARITY_CMD"
   }
 JSON
 }
@@ -611,6 +628,13 @@ export ADCIRC_BINS='$ADCIRC_BINS'
 export ADCSWAN_BINS='$ADCSWAN_BINS'
 export SWAN_UTIL_BINS='$SWAN_UTIL_BINS'
 EOMETA
+  # Singularity support requires there be a SIF file
+  # defined in the ADCIRC metadata file in $SCRIPTDIR/.adcirc-meta
+  if [ -n "${ADCIRC_SINGULARITY_SIF}" ]; then
+    cat <<EOMETA >> $ADCIRC_META_FILE
+export ADCIRC_SINGULARITY_SIF='$ADCIRC_SINGULARITY_SIF'
+EOMETA
+  fi
 }
 
 #
--- a/cloud/general/DOT-asgs-brew.sh
+++ b/cloud/general/DOT-asgs-brew.sh
@@ -208,6 +208,41 @@ goto() {
   esac
 }
 
+unload() {
+  CHOICES=();
+  case "${1}" in
+    adcirc)
+      # Save original PATH
+      local orig_path=$PATH
+
+      # Grab the first directory (up to first :)
+      local first=${orig_path%%:*}
+
+      # Everything after the first :
+      local tmp=${orig_path#*:}
+
+      # Grab the second directory (up to next :)
+      local second=${tmp%%:*}
+
+      # Everything after the second :
+      local rest=${tmp#*:}
+
+      if [ -x "$second/adcirc" ]; then       # adcirc was already in PATH
+        export PATH=${rest}
+        export ADCIRC_SINGULARITY_SIF=         # important because qscript.pl uses it to decide to do singularity stuff
+        echo "ADCIRC paths removed from PATH."
+        save profile ${_ASGSH_CURRENT_PROFILE}
+      else                                   # adcirc was not yet included in PATH
+        PATH=${SWANDIR}:${ADCIRCDIR}:${PATH}
+        echo "ADCIRC paths not found. PATH was not modified."
+      fi
+      ;;
+    *)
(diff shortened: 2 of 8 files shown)


wwlwpd:
@jasonfleming I've made some progress and I think I've got something running on QBD, but we need to work out what to do with the executable utilities that are not currently run in a queue script. The utilities are all currently in the `PATH` and I removed `ADCIRCDIR` from the calls to the utilties, but this was a hack to get things working.

Options:

# move all calls to any ADCIRC related executable to a queue script (I don't think this is feasible or a good idea)
# pair a local build with a containerized ADCIRC

Since putting everything in the queue call is not feasible, the process of adding ADCIRC that has a containerized component I recommend the following process:

- run `build adcirc`
- select version that is containerized
- ask if there is a container (if NO, proceed as is normal)
- if YES, ask what the `CONTAINER_CMD` is, which includes the path to the SIF file (initially Singularity eventually Docker)
- then build only the utilities,
  - `adcprep`
  - `hstime`
  - `unhcat.exe`
  - `HottifySWAN.x`
  - `adcirc` (needed for `adcirc -v`)
  - anything else not `padcirc`, `adcswan`, or `padcswan`
- make sure `PATH` is set correctly in ASGSH and in the container (Feng made sure all ADCIRC binaries were in the `PATH` on the container, so care in creating the container environment must be made)

We probably need to talk more about this and the different options, but this is to get around the fact that we can run anything in the container from the head node.

Any prelimary thoughts?

wwlwpd:
```
+-------------------------------------------------------------------+
| +---------------------------------------------------------------+ |
| | +-----------------------------------------------------------+ | |
| | |                                                           | | |
| | |                                                           | | |
| | |                                                           | | |
| | |                                                           | | |
| | |  Locally Available               Available on Container   | | |
| | |     adcirc (for -v)                 adcprep               | | |
| | |     adcprep                         adcirc                | | |
| | |     hstime                          padcirc               | | |
| | |     HottifySWAN.x                   adcswan               | | |
| | |     unhcat.exe                      padcswan'             | | |
| | |     ... ?                                                 | | |
| | |                                * via CONTAINER_CMD        | | |
| | |                                                           | | |
| | |                                                           | | |
| | |                                                           | | |
| | |                                                           | | |
| | |                                                           | | |
| | |                                                           | | |
| | |                                                           | | |
| | |         Loaded ADCIRC Utilities                           | | |
| | +-----------------------------------------------------------+ | |
| |                                                               | |
| |                                  Loaded Profile ENV           | |
| +---------------------------------------------------------------+ |
|                                                                   |
|                  ASGS Shell Environment                           |
+-------------------------------------------------------------------+
```

wwlwpd:
@jasonfleming good thoughts, thank you. Currently `CONTAINER_CMD` is associated with the ADCIRC version itself. But it can be broken up into to parts; `CONTAINER_CMD` could go into the platform's `info.sh` and we can simply define `ADCIRC_CONTAINER` (the `.sif` file). However, since testing with a Docker environment is not even a real need and a huge unknown, I'd want to leave it for just the singular use case.

What is the MVP? What do we need to get it TXLA22a running on QBD? Of the things you mentioned, what should we worry about doing here to meet the MVP?

`adcirc -v` can easily be replaced with `ADCIRC_VERSION` being set once in the ADCIRC wizard's metadata; but it's safer and much less work to just let ADCIRC build as it exists now. So I see anything that actually runs locally (i.e., `adcirc -v`) as fine. I don't think the MVP here is running "purely" via the container since that is not needed.

There is the possibility of incompatibility in data (maybe?) formats when it comes to what's read and/or written by the tools run on the headnode - I am seeing some odd behavior with my test run on QBD. I will look and address that with you separately. The forecasts do not seem to be getting posted, only the forecasts. Numerically, it seems to look "normal" to me from inspecting the timestep summaries printed to the log.

Please let me know what you consider "MVP" here. I think braking `CONTAINER_CMD` into 2 different things is a good idea.

- `CONTAINER_CMD` should be in the `patches/ADCIRC/$VERSION/info.sh`
- `ADCIRC_CONTAINER` (or `ADCIRC_SIF` since we're only testing singularity here) should be entered during the `adcirc build` process

What else?

wwlwpd:
> A lot of food for thought here. I think the MVP (or the most basic task list) is to do the following:
> 
>1. Create a `deploy adcirc` command in `asgsh` that is analogous to the `build adcirc` command, except in the case of `deploy` it would download the image and/or do whatever is necessary so that ASGS could use it. All ADCIRCs that were either built or deployed would show up in the output of `list adcirc`.
> 

The singularity container images need to be deployed already by the admins, as far as I can tell. The current use case is I simply know the location of the `.sif` file ahead of time. There are also some caveats, such as: all needed utilities need to be in the `PATH` of the effective user issuing the command via `singularity run` (it is like `docker exec`). So there is no deploying of images for ASGS or the user to do in our case here.

Instead it'll be an addition question during `build adcirc`. It'll simply ask if there is an associated singularity containter (`.sif`), if YES, then provide the full path. This is what will be set as `ADCIRC_CONTAINER` or probably `ADCIRC_SIF` since Docker is out of scope. The current changes in `qscript.pl` will ensure that the right `CMD` gets set in the queue SCRIPT (_Slurm_ since this is QBD specific for now).

To avoid confusion, I may even restrict this dialog to version 55* on QBD. People are not banging down the doors to run with singularity (I don't think it is very popular anywhere), so this will open up the conversation for people who need it elsewhere.

>
>2. The ASGS should know if the loaded adcirc has been built or deployed and should know its version string from `adcirc -v` and also be able to invoke it successfully in a queue script. This enables us to mix and match built and deployed ADCIRCs, even if they are the same exact version (actually the most common use case at this early stage).

Can you clarify this point more? I am not sure I understand.


> 3. Perform whatever light refactoring is required to make sure that we don't depend on running 
> any compiled ADCIRC codes or utilities on the head node. As I said above, my understanding is that 
> the ADCIRC utilities we run on the head nodes now are there more for convenience than necessity. 
> Still need to confirm that there won't be unpleasant surprises though. :-)

Ok, I think from the above you're ok with the MVP being essentially the current PR with the addition that I move `CONTAINER_CMD` to the platform and update `init-adcirc.sh` to ask for a `.sif` if it is set (where it stores it as `ADCIRC_CONTAINER`)?

Some caveats:

- I might make it `SINGULARITY_CMD` and `ADCIRC_SIF`.
- the local version of ADCIRC will still build and the user will have to make sure the version matches

I am okay with this, while incomplete it gets us running TXLA22a on QBD with 55* and leaves the field wide open to see how people may want to use it, if so. We can also claim we "added initial Singularity support' - that may invite others to try it and give us feedback.

Also in my previous post I mispoke:

> The forecasts do not seem to be getting posted, only the forecasts.

should be,

> The forecasts do not seem to be getting posted, only the **nowcasts**.

Also it's using the containerized `adcprep` and `padcirc` - `adcirc -v`, `hstime`, and `adcirc -v` are done locally. I will examine there runs and ping you outside this PR unless it's relevant. Functionally, the daily forecast has been running for a few days now.

wwlwpd:
Notes for the current push:

- using `ADCIRC_SINGULARITY_SIF` (will be written to metadata file in `./adcirc-meta` by `init-adcirc.sh`
- using `ASGS_SINGULARITY_CMD` - defined in `platforms/queenbeeD/info.sh`
- fixed long standing issue of ADCIRC paths accumulating in `PATH` (now they are added/removed, no more accretion)
- added `unload adcirc` command
- both `load adcirc`/`unload adcirc` reset `ADCIRC_SINGULARITY_SIF` (`load` does this before loading the "new" ADCIRC profile)

To do:

- [ ] document `unload adcirc` in `help`
- [ ] update `init-adcirc.sh` to ask about the "SIF" file if the platform has `ASGS_SINGULARITY_CMD` defined

wwlwpd:
@jasonfleming this is configured to work on QBC, QBD, SuperMikeIII, & supermic

1. build branch `issue-1588`
2. build ADCIRC v55.01*
3. when it asks you for a `.sif`, enter the path (see options below)
4. run a case that uses this version (I ran TXLA22a on all 4 machines successfully)

One all machines, the path is on the shared file systems at `/project/containers/images/adcirc.5bc04d6.rky8.5.sif`.

wwlwpd:
A few additional changes that happened here in addition to the singularity support:

1. I made `load adcirc` clean, meaning it only ever has one set of ADCIRC paths in `PATH`
2. I added `unload adcirc` that removes ADCIRC paths from `PATH`
3. on SuperMikeIII, netCDF4 needed to be built with some xml config lib, so I addusted the `./configure` slightly
4. added `unload adcirc` to `help` output
5. improved MPI detection (I think) so it only reaches for OpenMPI in the correct cases - found an edge case on one of the QueenBees

PR #1608, opened 2025-12-10
Fixed OpenMPI/MPI detection regression

wwlwpd:
Issue 1601: Detection was working before the change that broke this, but the change was an attempt to more clearly check for the different' cases. There are currently 4 cases for which we will skip the building of OpenMPI,

  - the compiler is Intel oneAPI and the expected MPI utils are found
  - the compiler is Inten "classic" and the expected MPI utils are found
  - the compiler is gfortran, and mpif90, etc are found in PATH
  - else, mpif90, etc are found already built in ASGS' opt/bin directory

Resolves #1601

### DIFF
--- a/cloud/general/asgs-brew.pl
+++ b/cloud/general/asgs-brew.pl
@@ -754,14 +754,35 @@ sub get_steps {
                 my ( $op, $opts_ref ) = @_;
                 my $bin          = qq{$asgs_install_path/$asgs_compiler/bin};
                 my $ok           = 1;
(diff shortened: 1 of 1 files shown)


wwlwpd:
A quick way to test is with the command ` ./init-asgs.sh -b -x "--run-steps openmpi"` in a freshly cloned out repository.

- where we expect Intel compilers, it won't try to build
- where we already have OpenMPI in gfortran environment, it won't try to build
- where we already have installed OpenMPI in a gfortran environment, it won't try to build
- where we are on a gfortran environment with no OpenMPI, it will download and build

PR #1611, opened 2025-12-11
Added netCDF4 build flags to eliminate need for libcurl

### LINKED ISSUE #1610: netCDF `./configure` needs some flags to disable "remote" stuff that requires dependencies we don't need
```
--enable-netcdf-4 --disable-dap --disable-byterange --disable-remote-tests
```

All the disable flags refer to remote access - `dap` also refers to OpenDAP/THREDDS server stuff.

Currently the change is in #1609, and may just keep it there. It is commit c31366f08e23e19c0e3fe3d4738241a3031119ae.

wwlwpd:
Issue 1610: the recent version of netCDF4 includes support for remote operations that we do not require and that limit portability on more restrictive platforms across non-GNU compilers (e.g., Intel). This commit adds build flags to disable these features.

Resolves #1610.

### DIFF
--- a/cloud/general/init-netcdf4.sh
+++ b/cloud/general/init-netcdf4.sh
@@ -67,7 +67,7 @@ if [ ! -e $OPT/bin/nc-config ]; then
   tar zxvf v${NETCDF4_C_VERSION}.tar.gz
   cd netcdf-c-${NETCDF4_C_VERSION}
   make clean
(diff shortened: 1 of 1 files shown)


wwlwpd:
This has been tested on QBD and the restrictive RHEL environment present on WWOD.

Issue #1613, opened 2025-12-22
generate HTML summary index.html for each upload from opendap_post2.sh


oodler577:
It would be easy and extremely helpful to generate a user friendly `index.html` with run summaries, links, and other infos so that when the files get uploaded. Scope is just an initial summary page, making the page more function can be part of a future story.

<img width="1824" height="756" alt="Image" src="https://github.com/user-attachments/assets/5f69b569-18cf-4e7e-9672-b860dc5fd25e" />

e.g.,
https://fortytwo.cct.lsu.edu/thredds/fileServer/singularity/2025/GFS/2025122018/TXLA22a/supermic.hpc.lsu.edu/TXLA22a_GFS_smic_be/nowcast/index.html

<img width="1113" height="986" alt="Image" src="https://github.com/user-attachments/assets/742aaf16-d810-46cd-8d12-0b1f522c6247" />

example command,

```
echo "<h1>Output Files</h1>" > ./index.html;
ls -1 | awk 'BEGIN{print "<ul>"} {printf "<li><a href=\"%s\">%s</a></li>\n",$0,$0} END{print "</ul>"}' >> ./index.html 2>&1;
echo "<h1>run.properties</h1><pre>" >> ./index.html; cat ./run.properties >> ./index.html;
echo "</pre>" >> ./index.html
```
but I'd like to use a template and make it much more useful...

jasonfleming:
This is a fantastic idea!! Would the html summary be posted to the same server or a different one (that would be for a more limited audience)?

wwlwpd:
> This is a fantastic idea!! Would the html summary be posted to the same server or a different one (that would be for a more limited audience)?

It could just be HTML generated for each run, then uploaded with the rest; if it's on a webserver like fortytwo, then `index.html` will be loaded automatically by the browser.

wwlwpd:
@jasonfleming I began looking at this, we have `postAdditionalFiles` that we can use to add `index.html` that gets generated by the script can inspect file - question is, where should this script be run? E.g., should it be done as part of an existing hook or embedded elsewhere? I think a hook may be the right place, so which one?

It seems like it should be after the quality check passes, is there a hook right before that? What do you think?

jasonfleming:
[...]

PR #1618, opened 2026-01-08
Adding script for HTML generator of `index.html` for more effective use of browser based results

wwlwpd:
#1613


wwlwpd:
This was resolved with #1677.

PR #1619, opened 2026-01-09
Updated patches to remove mpif90 for when compiler is Intel, added flag for METIS

### LINKED ISSUE #1569: remove use of `mpif90` from all HPCs that prefer Intel for building ADCIRC
Intel never actually provided `mpif90`, this has always been an MPICH/OpenMPI thing. Intel has provided `mpiifort` for a very long time.

wwlwpd:
Issue 1569: cleaned up the Intel section so that mpif90 is no longer used, for Intel classic mpiifort is the default; for oneAPI it is mpiifx.

Versions 55.02, 55.01*, and v53-release were not compiling under oneAPI, this commit also includes the addition of a compiler flag that allows for the implicit functions used by the version(s) of METIS used by these ADCIRC versions. METIS only breaks for these version with icx (oneAPI) - there is no issue with icc (Intel's "classic" C compiler).

Resolves #1569.

I need to test this under GCC and Intel classic:

- [x] (VERIFIED) test builds on Debian / Ubuntu
- [x] (VERIFIED) test builds on supermic (where `ifort`/`mpiifort` are available)


jasonfleming (approved):
Looks great @wwlwpd , thank you, please merge. :-)

PR #1624, opened 2026-01-28
Fixing metadata issues

### LINKED ISSUE #924: Dubious WindModel entry
When attempting to post-process a retrospective analysis of a historical storm,  the `run.properties` file has the entry `WindModel : vortex-nws0`, I am assuming this hindcast uses GAHM+SWAN, therefore shouldn't the entry be `vortex-nws320`? Is this a typo?

### LINKED ISSUE #1617: track control node hostname and `asgs_main.sh` procid in `run.properties`
This came about because I just figured out where my ASGS instance running on SuperMIC for the last month, since 2025120500 successfully. I was on smic1 but was on smic2 - so 2 things to come out of this are, 1) the singularity container approach is extremely stable! and 2) I think we need to track the hostname of the controller (where asgs_main.sh is running) and the process id of asgs_main.sh in run.properties.

This could be wrapped up with #1613 since this information could be used to generate the HTML summaries for quick views.

### LINKED ISSUE #1622: Correct the writing of properties to `run.properties` file on nowcasts
The basic/general properties for a particular scenario are written to `$RUNDIR/run.properties` before polling for new nowcast data, and then this file is moved to the scenario directory when new nowcast data have been obtained. This does not seem to be working as expected for synoptic meteorological forcing scenarios (at least). This will interfere with any subsequent script that relies on metadata from `run.properties`.

jasonfleming:
resolves #1617 resolves #924 resolves #1622

### DIFF
--- a/asgs_main.sh
+++ b/asgs_main.sh
@@ -1301,10 +1301,10 @@ monitorJobs()
       fi
    fi
    # record memory high water mark to the run.properties file; this
-   # can be used for resource monitoring and in troubleshooting out 
+   # can be used for resource monitoring and in troubleshooting out
    # of memory issues
    if [[ $QUEUESYS == "SLURM" ]]; then
-      echo "hpc.job.${JOBTYPE}.$(<jobID).sacct.maxrss.bytes : $(sacct -j $(<jobID).batch --format=MaxRSS --noconvert --noheader)" >> run.properties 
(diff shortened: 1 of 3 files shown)


jasonfleming:
Tested successfully. Please review and merge if approved @wwlwpd. :-)

PR #1641, opened 2026-02-20
minimal changes needed to support ntasks method of increasing memory

wwlwpd:
This is to access more memory per node for `serial` tasks the "right way" in Slurm- starting with supermic.

### DIFF
--- a/bin/guess
+++ b/bin/guess
@@ -27,7 +27,9 @@ _guess_platform()
 # !!!
 # add new platforms to platforms/platforms-lookup.txt
 # !!!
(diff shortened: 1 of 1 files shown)


wwlwpd:
And I think before this is all said and done, we'll go ahead and remove support for `--mem` or `--mem-per-task`, LSU HPC indicated that this was not the correct way to increase memory. But to increase `--ntasks` to multiply the amount of RAM they give to each CPU. In the case of supermic, it's 4GB per core.

jasonfleming (commented):
Was the removal of queenbeeC from `bin/guess` intentional?

Issue #1635, opened 2026-02-12
Create workaround for out-of-memory issues for serial jobs


jasonfleming:

wwlwpd:
We should update `qscript.template` to use Templete Toolkit fully (replace custom rendering loop) in a separate issue, next.

Issue #1354, opened 2024-07-28
Enable `control_file_gen.pl` to populate namelist parameters individually 


jasonfleming:
The availability and effect of namelist parameters has changed for different ADCIRC versions. Not all parameters should be populated with default values for a given namelist if the version of ADCIRC used with that fort.15 file does not support them. Currently, `control_file_gen.pl` populates all namelist values for each namelist for all versions of ADCIRC. The search-and-replace should be more granular so that only the parameters in the template are populated. Related to #228.

wwlwpd:
@jasonfleming - I'd like to discuss beginning this sooner rather than later; perhaps e can begin by simply exposing existing variables (parameter packages) in the default `mesh_defaults.sh` for those meshes with properly templated `fort.15` files?

I saw you already have a branch, issue-1354 - check this short POC from a test branch I created (just as an example) - [f976675](https://github.com/StormSurgeLive/asgs/commit/f976675072161bdb1c6427e0bcb9c2788dff42d9)

PR #1307, opened 2024-05-11
Fixed ppn hint and jobID capture on mike

jasonfleming:
resolves #1306 with additional testing now

Issue #1306, opened 2024-05-10
Correct `--ntasks-per-node` on mike


jasonfleming:
I think they have made a configuration change ... in the `padcirc.slurm` file for the Wind10m scenario, which normally runs on 16 cores, the `--ntasks-per-node` was set to 64 (which apparently used to work) but now it needs to be set the same as `--ntasks` ... I will be logging an issue for this, fixing the `HPC_PPN_Hints()` function in `platforms.sh` and restarting it...

jasonfleming:
There is also an issue with the `scenario.status.json` file that is getting generated on mike. Each job is listed twice.

PR #1643, opened 2026-02-21
Adding SuperMikeIII to the list of platforms where SERQUEUE_NTASKS is supported

wwlwpd:
Issue 1635: Adding SuperMikeIII, this also allows SERQUEUE_NTASKS for SuperMikeIII to be updated in the ASGS configuration file being used in the situation that the OOM is happening.

Related to #1635.

### DIFF
--- a/platforms.sh
+++ b/platforms.sh
@@ -383,7 +383,7 @@ HPC_PPN_Hint()
    ;;
    "qbd.loni.org")
      if [[ "$QUEUEKIND" == "serial" ]]; then
-       echo $SERQUEUE_NTASKS # this is defined in ./platforms/queenbeeQ
+       echo $SERQUEUE_NTASKS # this is defined in ./platforms/queenbeeD
      else
        echo $DEFAULT_PPN
      fi


platforms.sh
```diff
@@ -383,14 +383,14 @@ HPC_PPN_Hint()
...
+       echo $SERQUEUE_NTASKS # this is defined in ./platforms/queenbeeQ
      else
        echo $DEFAULT_PPN
      fi
    ;;
    "mike.hpc.lsu.edu")
-     if [[ "$QUEUENAME" == "single" && "$CPUREQUEST" -lt 64 ]]; then
-       echo $CPUREQUEST
+     if [[ "$QUEUEKIND" == "serial" ]]; then
+       echo $SERQUEUE_NTASKS # this is defined in ./platforms/mike-lsu
```
jasonfleming: In `./platforms` in this branch I see `mike` but not `mike-lsu` ... is there a plan to disambiguate `mike` in `./platforms` or is this just a typo in the comment here

platforms/mike/init.sh
```diff
@@ -17,6 +17,7 @@ export QUOTACHECKCMD=null
 export ALLOCCHECKCMD=null
 export QUEUENAME=workq
 export SERQUEUE=single
+export SERQUEUE_NTASKS=3 # 12G for slurm, $SERQUEUE_NTASKS * 4G, applied to --ntasks
```
jasonfleming: Do we know whether this will work given the previous use of `CPUEQUEST` always being 64 on mike in the single queue? Or whether this approach will be significantly better (at least it is consistent with the other platforms)?

platforms.sh
wwlwpd: Yeah, that's a little forward thinking. It'll make sense after we merge #1354.

platforms/mike/init.sh
wwlwpd: @jasonfleming this is what I was asking about in Slack. Sam & Feng said MikeIII works exactly like all the other machine. So I think our change is doing essentially the same thing, but allow us to control available memory it using `SERQUEUE_NTASKS` both by default and in the ASGS config if we need to increase it. We can discuss further next time we circle up. There is no rush here as far as I know, it was just for completeness.

wwlwpd: After some digging, see 72fae1d0f and PR #1307 and Issue #1306.

PR #1656, opened 2026-03-19
Adding support for ADCIRCS v56.2.1

wwlwpd:
Need to test. @jasonfleming note that there is a necessary patch for `src/adcirc.F` in v56.1.0's patch-set; v56.2.1 doesn't have it, presumably this was one of the things they fixed. However I didn't yet look at my fix versus theirs.

We need to test these on all of our supported machines, I will keep it a draft until we figure out what to do about the `adcirc.F` patch for v56.1.0.

Testing needs to be done across a representative sampling machines and all compilers (intel classic, intel oneAPI, GCC).


jasonfleming:
Thank you @wwlwpd based on the need for this extra patch and some of the issues that have been fixed in v56.2.1 vs v56.1.0, my recommendation is to skip v56.1.0 and focus this effort on v56.2.1.

wwlwpd:
@jasonfleming I removed v56.1.0, so this only adds v56.2.1 - I am building this now, but I think it is ready for review.

jasonfleming (approved):
Looks excellent thank you!

PR #1661, opened 2026-03-23
Integrating asgs-lint into asgs_main.sh startup

### LINKED ISSUE #1603: `asgs-lint` added to initial `run` command
A good suggestion was to run asgs-lint when loading a profile with a defined configuration file, run automatically and report issues if they are found and offer to continue anyway (Operator can hit enter to accept the default “Yes”)

wwlwpd:
Issue 1603:

Resolves #1603.

### DIFF
--- a/asgs_main.sh
+++ b/asgs_main.sh
@@ -1661,7 +1661,8 @@ handleFailedJob()
       fi
    fi
 }
-#
+
+# import functions
 source $SCRIPTDIR/monitoring/logging.sh
 source $SCRIPTDIR/platforms.sh            # this includes source $SCRIPTDIR/monitoring/logging.sh
 source $SCRIPTDIR/properties.sh           # read properties file into a hash
@@ -1744,9 +1745,10 @@ fi
 readConfig # now we have the instancename and can name the asgs log file after it
 setSyslogFileName     # set the value of SYSLOG in monitoring/logging.sh
 nullifyHooks          # in manageHooks.sh
+addDefaultHooks       # after hooks are nullified, add back defaults
 #
 consoleMessage "$I START_INIT $GRIDNAME $HPCENVSHORT"
-executeHookScripts "START_INIT"
+executeHookScripts "START_INIT" || exit $?
 #
 # set a trap for a signal to reread the ASGS config file
 trap 'echo Received SIGUSR1. Re-reading ASGS configuration file. ; readConfig' USR1
--- a/cloud/general/asgs-brew.pl
+++ b/cloud/general/asgs-brew.pl
@@ -743,6 +743,8 @@ sub get_steps {
                 ASGS_MAKEJOBS          => { value  => qq{$makejobs},                              how  => q{replace} },          # passed to make commands where Makefile supports
                 ASGS_MESH_DEFAULTS     => { value  => qq{$scriptdir/config/mesh_defaults.sh},     how  => q{replace} },          # list of supported meshes
                 ASGS_PLATFORMS         => { value  => qq{$scriptdir/platforms.sh},                how  => q{replace} },          # list of supported platforms
+                QUALITYSETTING         => { value  => q{strict},                                  how  => q{replace} },          # default quality check level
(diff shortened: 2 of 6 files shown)


wwlwpd:
@jasonfleming I decided to make `LINTSETTING=warn` by default, which is non-blocking. It is so we don't break working, but "nonconforming" configurations for no reason.

wwlwpd:
@jasonfleming I did notice that `removeScriptFrom` does not work when used from the ASGS_CONFIG - there's some ordering with the `nullifyHooks` that seemed awkward, but I made it work here. Though `asgs-lint` is always used, it can be ignored and mostly hidden using `LINTSETTING`.

wwlwpd:
@jasonfleming ready for review

jasonfleming:
building this now for testing

jasonfleming:
I tested this and found that `asgs-lint` only ran as expected when I executed `run` from `$SCRIPTDIR`. For example, if I am in `$SCRATCH` and type `run` I get the following error:
```
(!! warning) asgs_main->manageHooks->addScriptTo_START_INIT(): The 'START_INIT' hook script 'lint-checks.sh' was not found or is not executable, so it will not be used.
```

wwlwpd:
> I tested this and found that `asgs-lint` only ran as expected when I executed `run` from `$SCRIPTDIR`. For example, if I am in `$SCRATCH` and type `run` I get the following error:
> 
> ```
> (!! warning) asgs_main->manageHooks->addScriptTo_START_INIT(): The 'START_INIT' hook script 'lint-checks.sh' was not found or is not executable, so it will not be used.
> ```

Thanks @jasonfleming - just saw this, I will fix that. A lot of things assume `$SCRIPTDIR`, I personally never `run` outside of `$SCRIPTDIR`, but that's just out of habit.

wwlwpd:
> I tested this and found that `asgs-lint` only ran as expected when I executed `run` from `$SCRIPTDIR`. For example, if I am in `$SCRATCH` and type `run` I get the following error:
> 
> ```
> (!! warning) asgs_main->manageHooks->addScriptTo_START_INIT(): The 'START_INIT' hook script 'lint-checks.sh' was not found or is not executable, so it will not be used.
> ```

@jasonfleming I think this is a problem with the current incarnation of manageHooks.sh - execution assumes the script is in`$SCRIPTDIR` prepending it with `$SCRIPTDIR`. The most consistent course of action would be to do nothing here. I could add some code in the execution function to add `$SCRIPTDIR` if there is no leading path (relative or absolute).

wwlwpd:
@jasonfleming I fixed the issue of the hook not finding the right hookscript on `run` by fixing an inconsistency in `manageHooks.sh`, it was unconditionally prepending `$SCRIPTDIR` to `$hc`; so now this only happens if no path separator (`/`) is found in `$hs`.

Do you have any idea how I can test with other hooks to make sure it doesn't break something basic, or do you think this approach is pretty safe?

wwlwpd:
>     Default for LINTSETTING is 'warn', which is non-blocking. This is so
>     that currently working, but ASGS configs that are not fully conformed
>     to asgs-lint do not suddenly start to break. But a "!!! warning" is
>     displayed in the STDOUT so hopefully they can either address the warning
>     or alert upstream devs to a false negative (i.e., is a valid
>     configuration).
> 
>     A user can adjust LINTSETTING in:
> 
>     ~/.asgsh_profile # overrides the global ASGS default (currently, 'warn')
> 
>     in $ASGS_CONFIG  # overrides even the user's default in ~/.asgsh_profile
>

jasonfleming (dismissed):
Tested `LINTSETTING` successfully with `info`, `warn`, `prompt`, and `strict`. Did not notice any differences in behavior with `info` compared to `warn`, which is fine. Setting this to `off` did not stop asgs-lint from linting or silence the messages, which is probably fine also, because everything still ran. Thank you!

oodler577:
@jasonfleming, difference in `warn` and `info` is subtle:

image

vs

image

jasonfleming:
Thank you I missed that. :-)

wwlwpd:
I made a small formatting tweak since approval, so gonna merge with bypass.

PR #1670, opened 2026-03-30
Automatically count the number of forecast scenarios

### LINKED ISSUE #1666: `asgs-lint` make the scenario package section static scanning more robust
e.g.,
 
- fully verify `SCENARIOPACKAGESIZE` and cases actually found
- ...

jasonfleming:
I have tested this successfully with the included test config ; resolves #1666


oodler577:
> I have tested this successfully with the included test config ; resolves #1666

That's great. Is it totally ignoring `SCENARIOPACKAGESIZE` now? It would be surprising (and confusing) if that variable no longer does anything, so can we turn it effectively into an exception, that warns if the actual number of scenarios doesn't match the number defined in `SCENARIOPACKAGESIZE`, if it is defined?

Also, what should the `asgs-lint` check scan for now in this section? It was not robust to begin with, but it does take `SCENARIOPACKAGESIZE` and reports as a fatal error when there are not enough scenarios defined.

jasonfleming:
I have been thinking more about this, and I agree that it is confusing for `SCENARIOPACKAGESIZE` to become suddenly meaningless. How about this instead:

1. If `SCENARIOPACKAGESIZE` is set to an integer, it is checked by `asgs-lint` for consistency with the number of scenarios. It would emit a `warning` if there are more scenarios defined in the `case` clause than `SCENARIOPACAKGESIZE` and a fatal `error` if there are less. 
2. If `SCENARIOPACKAGESIZE=$(get-scenario-package-size)` then the ASGS counts them via a separate script `bin/get-scenario-package-size`. 
3. If `SCENARIOPACKAGESIZE` is not set in the ASGS config at all, then the number of defined forecast scenarios is counted by ASGS with the `bin/get-scenario-package-size` script.

oodler577:
Unless it is impossible, I would just warn (see where `${W}` used for visibility) that the numbers don't match and that it's running all that's defined in the `case` statement. It could also warn that `SCENARIOPACKAGESIZE` has become deprecated and it can be removed. The point is to make it clear that variable is going away. And I can turn that `asgs-lint` check to say the same thing. What do you think?

oodler577:
I can also update the template on the tools site to not use it.

wwlwpd:
What @oodler577 said 😆

jasonfleming:
Ok that makes sense. :-)

jasonfleming:
Tested successfully and ready for another review.

asgs_main.sh
```diff
@@ -2840,8 +2845,17 @@ while [ true ]; do
...
-   ENSTORM="forecast"
-   logMessage "$ENSTORM: $THIS: Starting forecast scenarios for advisory '$ADVISORY'."
+   # Count the number of forecast scenarios in the scenario package.
+   # Should this be a hook script attached to the START_FORECAST_STAGE
+   # hook?
+   THIS="asgs_main.sh"
+
+   logMessage "$THIS: Counting the number of scenarios in the scenario package."
+   numScenarios=$(get-scenario-package-size)
+   if [[ $SCENARIOPACKAGESIZE -ne $numScenarios ]]; then
```
wwlwpd: @jasonfleming should this check also only happen if `SCENARIOPACKAGESIZE` is set and is not `auto`?

jasonfleming: Thank you for flagging this. In the process of correcting it I have found some other issues in `asgs_main.sh` where `SCENARIOPACKAGESIZE` was still being used instead of the autocounted value.

jasonfleming:
Thank you @wwlwpd I have corrected the handling of `SCENARIOPACKAGESIZE=auto` in the ASGS config and added testing to cover it. Works as expected and ready for another review. :-)

wwlwpd:
You mentioned this issue in the code, it's been there since I converted all the "TODO"s in the code to issue, https://github.com/StormSurgeLive/asgs/issues/593

wwlwpd (approved):
Thank you!

PR #1694, opened 2026-05-03
Redirect chatter from `source $CONFIG` to `$SYSLOG` and remove duplicate properties from `run.properties`

### LINKED ISSUE #1692: Prevent duplicate properties from being written to `run.properties` for wind10m layer
The wind10m layer is generated as a separate model execution and its properties are written to the `run.properties` file even if they are the same as the full physics execution. This results in duplicate properties being written to the `run.properties` file. Duplicated properties should not be written to the file.

### LINKED ISSUE #1693: Fix erroneous automated counting of scenario package size caused by chatter from config file
When executing `source $CONFIG` to count the number of forecast scenarios in the scenario package, the results can be contaminated with chatter from `echo` statements in the config file.

jasonfleming:
Tested this successfully locally with vortex forcing. Next step is full scale test. resolves #1692, resolves #1693

### DIFF
--- a/asgs_main.sh
+++ b/asgs_main.sh
@@ -1641,11 +1641,12 @@ handleFailedJob()
    ARCHIVEBASE=${14}
    THIS="asgs_main.sh>handleFailedJob()"
    # check for duplicate properties in the run.properties file
-   numDuplicateProperties=$(awk 'BEGIN { FS=":" } a[$1]++ {dup++} END { printf "%d", dup}' run.properties)
+   # check for duplicate property keys in the run.properties file (trimmed keys)
+   numDuplicateProperties=$(awk -f $SCRIPTDIR/util/output/countDuplicatePropertyKeys.awk run.properties)
    if [[ $numDuplicateProperties -ne 0 ]]; then
-      warn "There are '$numDuplicateProperties' duplicate properties in the '$ADVISDIR/$ENSTORM/run.properties' file."
-      duplicateProperties=$(awk 'BEGIN { FS=":" } {count[$1]++} END {for (item in count) if (count[item] > 1) print item, count[item]}' run.properties)
-      logMessage "The duplicate properties and the number of times they were each duplicated are as follows: '$duplicateProperties'."
+      warn "There are '$numDuplicateProperties' duplicate property key occurrences in the '$ADVISDIR/$ENSTORM/run.properties' file."
+      duplicateProperties=$(awk -f $SCRIPTDIR/util/output/itemizeDuplicatePropertyKeys.awk run.properties)
+      logMessage "The duplicated property keys and the number of times they were found are as follows: '$duplicateProperties'."
    fi
    # check to see that the job did not conspicuously fail
    if [[ -e $ADVISDIR/${ENSTORM}/jobFailed ]]; then
--- a/asgs_main.sh
+++ b/asgs_main.sh
@@ -2855,7 +2855,7 @@ while [ true ]; do
    THIS="asgs_main.sh"
 
    logMessage "$THIS: Counting the number of scenarios in the scenario package."
-   numScenarios=$(get-scenario-package-size)
+   numScenarios=$(get-scenario-package-size $SYSLOG)
    # if the Operator did not set the SCENARIOPACKAGESIZE in the config file,
    # its value will be -1
    if [[ $SCENARIOPACKAGESIZE != "auto" && $SCENARIOPACKAGESIZE -ne $numScenarios && $SCENARIOPACKAGESIZE -ne -1 ]]; then
--- a/asgs_main.sh
+++ b/asgs_main.sh
@@ -378,7 +378,6 @@ prep()
     NAFILE=${17}  # full domain nodal attributes file
     #
     THIS="asgs_main.sh>prep()"
-    echo "time.adcprep.start : $(date +'%Y-%h-%d-T%H:%M:%S%z')" >> ${STORMDIR}/run.properties
     # set the name of the archive of preprocessed input files
     PREPPED=$PREPPEDARCHIVE
     if [[ $START = coldstart ]]; then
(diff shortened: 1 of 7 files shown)


bin/get-scenario-package-size
```diff
@@ -35,7 +35,7 @@ _getScenarioPackageSize() {
     # maintain backward compatibility with old config files
     SCENARIOPACKAGESIZE=-1
     ENSEMBLESIZE=-1
-    source $ASGS_CONFIG
+    source $ASGS_CONFIG > /dev/null 2>&1  # suppress chatter from the config file
```
wwlwpd: I know it is tempting, but we can't do this, it'll end up costing us hours in debugging time. What is the purpose of this?

generateDynamicInput.sh
```diff
@@ -317,7 +317,11 @@ generateDynamicInput()
...
-    if [[ -e $SCENARIODIR/wind10m.run-control.properties ]]; then
-       cat $SCENARIODIR/wind10m.run-control.properties >> $SCENARIODIR/run.properties 2>> $SYSLOG
-    fi
+    # copy properties that are specific to a certain
+    # layer to the general run.properties file
+    for l in ${layers[@]}; do
+        if [[ -e $SCENARIODIR/${l}.run-control.properties ]]; then
+            cat $SCENARIODIR/${l}.run-control.properties | grep $l >> $SCENARIODIR/run.properties 2>> $SYSLOG
+        fi
+    done
```
wwlwpd:
```suggestion
    for l in "${layers[@]}"; do
        if [[ -f "$SCENARIODIR/${l}.run-control.properties" ]]; then
            grep "$l" "$SCENARIODIR/${l}.run-control.properties" \
                >> "$SCENARIODIR/run.properties" 2>> "$SYSLOG"
        fi
    done
```

wwlwpd: Suggestions:

- Quoting: keeps things safe if paths or values ever contain spaces.
- No `cat`: `grep` reads files directly.
- `-f` vs `-e`: ensures you're dealing with a regular file.
- `"${layers[@]}"`: preserves each element correctly.

asgs_main.sh
```diff
@@ -1641,6 +1640,13 @@ handleFailedJob()
    JOB_FAILED_LIST="${13}"
    ARCHIVEBASE=${14}
    THIS="asgs_main.sh>handleFailedJob()"
+   # check for duplicate properties in the run.properties file
+   numDuplicateProperties=$(awk 'BEGIN { FS=":" } a[$1]++ {dup++} END { printf "%d", dup}' run.properties)
+   if [[ $numDuplicateProperties -ne 0 ]]; then
+      warn "There are '$numDuplicateProperties' duplicate properties in the '$ADVISDIR/$ENSTORM/run.properties' file."
+      duplicateProperties=$(awk 'BEGIN { FS=":" } {count[$1]++} END {for (item in count) if (count[item] > 1) print item, count[item]}' run.properties)
+      logMessage "The duplicate properties and the number of times they were each duplicated are as follows: '$duplicateProperties'."
+   fi
```
wwlwpd, line 1650: @jasonfleming is this supposed to be counting duplicate lines or duplicate properties/keys?

bin/get-scenario-package-size
jasonfleming: In the `$CONFIG` file, the automated counting of the scenario package makes the `echo` statement in the fall-through `*)` condition of the `case` statement in the scenario package configuration obsolete. That echo statement also contaminates the value returned from of `$(get-scenario-package-size)`. I'm not sure why/how it would impact debugging.

asgs_main.sh
jasonfleming: This is counting duplicate keys in the `run.properties` file.

bin/get-scenario-package-size
wwlwpd: > In the `$CONFIG` file, the automated counting of the scenario package makes the `echo` statement in the fall-through `*)` condition of the `case` statement in the scenario package configuration obsolete. That echo statement also contaminates the value returned from of `$(get-scenario-package-size)`. I'm not sure why/how it would impact debugging.

It's just hiding an error, that's never worked out well in my experience. What's the root cause of the error it is hiding? We can discuss next time we meet so I understand what the error is.

asgs_main.sh
wwlwpd: I will DM you.

bin/get-scenario-package-size
wwlwpd: I should just ask, what error is the config file throwing when `source`d?

jasonfleming: It is not hiding an error. As I said, it is disposing the known and expected output from the  `echo` statement in the `$CONFIG` file that tells the Operator that the scenario package counter has exceeded the number of scenarios. Which is something the script is intentionally doing as a way of counting the number of forecast scenarios. If it helps, I can capture and report `stdout` and `stderr` from the `source $CONFIG` command to `$SYSLOG`. In any case, `asgs_main.sh` re-reads the `$CONFIG` file with `source $CONFIG` at the start of every cycle and does not redirect `stdout` or `stderr` so I'm not sure what information we're losing when we do this in `get-scenario-package-size`.

asgs_main.sh
jasonfleming: Good catch. I have improved the awk scripts by removing white space from the keys before checking for duplicates. Will continue to test in live operations.

bin/get-scenario-package-size
jasonfleming: I've redirected this message to `$SYSLOG` with a (hopefully reassuring) log message indicating that it is not an error.

oodler577: Ok. I think the issue is that we don't want to suppress `STDERR` but can suppress `STDOUT`, so if you wanted to do `> /dev/null` that's fine, but `> /dev/null 2>&1` is not since that's basically merging `STDERR` (`2`) with `STDOUT` (`1`).

wwlwpd:
@jasonfleming I'll do the final review when you put it in Ready for review. We probably already covered everything.

bin/get-scenario-package-size
jasonfleming: ok

jasonfleming:
Tested successfully and ready for review.

wwlwpd (approved):

PR #1695, opened 2026-05-06
Adding support for bc if it is not found on the platform

### LINKED ISSUE #1691: Check for availability of `bc` during installation
This utility is not necessarily available on all platforms by default.

wwlwpd:
Issue 1691: bc and dc (same tarball) will be installed during ASGS' install if it is not found on the system. It is present on most machines, but modern Linux distros seem to be removing older utilities.

Resolves #1691.

### DIFF
--- a/cloud/general/asgs-brew.pl
+++ b/cloud/general/asgs-brew.pl
@@ -1150,6 +1150,31 @@ sub get_steps {
             },
         },
         {
+            key         => q{bc},
+            name        => q{Step for installing bc and dc},
+            description => q{Install bc for shell-based floating point calculations},
+            pwd         => qq{$scriptdir},
+            command     => qq{bash ./cloud/general/init-bc.sh $asgs_install_path},
+            clean       => qq{bash ./cloud/general/init-bc.sh $asgs_install_path clean},
+            skip_if     => sub {
+                local $?;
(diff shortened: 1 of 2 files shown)


wwlwpd:
This is hard to test since it won't build if `bc` is present on the system via `which` - meaning it's in `PATH`. We could just always build it if not in `$SCRIPTDIR/opt/bin`

jasonfleming (approved):
I have tested this successfully by introducing a typo into `asgs-brew.pl` to make it check for a nonexistent executable: 
```perl
system(qq{abc --version > /dev/null 2>&1});
```
and found that `bc` was built and installed successfully:
```bash
> which bc
/home/jason/Campaigns/Development/2026/issue-1691/opt/bin/bc
```
Thank you!

wwlwpd:
> I have tested this successfully by introducing a typo into `asgs-brew.pl` to make it check for a nonexistent executable:
> 
> ```perl
> system(qq{abc --version > /dev/null 2>&1});
> ```
> 
> and found that `bc` was built and installed successfully:
> 
> ```shell
> > which bc
> /home/jason/Campaigns/Development/2026/issue-1691/opt/bin/bc
> ```
> 
> Thank you!

Nice! That's basically what I was doing, too.

PR #1701, opened 2026-05-11
Add option to `fetch` to force using ssh to clone

### LINKED ISSUE #1700: add `--force-ssh` to `bin/fetch`
Need to be able to force git@ ssh-based urls even in releases for dev and operator convenience.

wwlwpd:
Issue-1700: When using a "release" tar.gz or zip, bin/fetch will fallback to using the https protocol for cloning supported repositories; this commit adds an option, --force-ssh, to force ssh; if for some reason a user needs to force https, there is also the option to --force-https.

Example,

  asgs-shell>fetch cera # may default to https authentication

so instead, do:

  asgs-shell>fetch cera --force-ssh # now uses the ssh key that should be present

Resolves #1700.

### DIFF
--- a/bin/fetch
+++ b/bin/fetch
@@ -1,5 +1,32 @@
 #!/usr/bin/env bash
 
+ITEM=${1}
+
+shift
+
+FORCE_SSH=0
+FORCE_HTTPS=0
+
+# parse remaining optional flags
+while [[ $# -gt 0 ]]; do
+  case "$1" in
+    --force-ssh)
+      FORCE_SSH=1
+      ;;
+    --force-https)
+      FORCE_HTTPS=1
+      ;;
+    *)
+      ;;
+  esac
+  shift
+done
+
+if [[ $FORCE_SSH -eq 1 && $FORCE_HTTPS -eq 1 ]]; then
+  echo "--force-ssh and --force-https are mutually exclusive"
+  exit 1
+fi
+
 if [ -z "$_ASGSH_PID" ]; then
   SCRIPTDIR=$2
   if [ -z "$SCRIPTDIR" ]; then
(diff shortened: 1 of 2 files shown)


wwlwpd:
I added another root that makes bouncing between ASGS tar.gz and git repo much easier to tweak things so they work well in the exported environment as well as the .git environment. `regit` - makes it a git repo again.

jasonfleming:
This looks good but I don't fully understand it because I've never had an issue with the `fetch` command. So I want to understand the issue that other Operators are experiencing. I always use the `git clone https://github.com/xxxx` syntax to download the repo. What was the unexpected behavior that this resolves?

wwlwpd:
> This looks good but I don't fully understand it because I've never had an issue with the `fetch` command. So I want to understand the issue that other Operators are experiencing. I always use the `git clone https://github.com/xxxx` syntax to download the repo. What was the unexpected behavior that this resolves?

`fetch` auto-detects the protocol that the `git` remote uses (e.g., `https` or `git@` being _ssh_ based); but sometimes I need to use one of the managed repos (asgs-lint, asgs-mon, etc) to use a different protocol so that I can push to it - or pull a restricted one - more conveniently). I run into it, maybe uniquely, when testing the .tar.gz releases and setting up WSL2 images.

Basically when I need to use git for a supported repo in `fetch`, in a way different than how it is used (or not used) in the current asgs SCRIPTDIR.

jasonfleming:
Ok I think I understand now. In order to push to our public repo, or to pull from a private repo, you need authentication with ssh keys, and the default https protocol won't work for that. Is that right?

Also, what does the `regit` command do?

wwlwpd:
> Ok I think I understand now. In order to push to our public repo, or to pull from a private repo, you need authentication with ssh keys, and the default https protocol won't work for that. Is that right?
> 
> Also, what does the `regit` command do?

It is something I have needed for a while in this same vein, if I am working from a .tar.gz (not a git repo), it let's me force it to be a git repo so that I can push any changes I make in place to an upstream branch.

jasonfleming:
Ok I see, so `regit` turns a bundle of ASGS code into a freshly initialized git repo. Then it relates it to the origin repo based on the `VERSION` sha found in the code bundle and finally creates a local branch from that sha. Is that correct?

wwlwpd:
> Ok I see, so `regit` turns a bundle of ASGS code into a freshly initialized git repo. Then it relates it to the origin repo based on the `VERSION` sha found in the code bundle and finally creates a local branch from that sha. Is that correct?

You got it - and it works as long as I do not forget to update what is in the `./VERSION` file. Though we don't have a `development` branch, I think I need to fix that. I will keep _development_ but consider it `master` just to keep in place our long standing naming conventions. Stand by.

jasonfleming (approved):
Looks great, and I understand it now. :-)

PR #1706, opened 2026-05-12
Formalizing v53.05.live.0 as a separate LTS version.

### LINKED ISSUE #1705: formalize v53.05.live.0 patchset
v53release contains custom patches and fixes, we are doing for it what we have done with v56.0.4.live; keep v53release around for a while even though this is identical; the only difference is that that .live version is pulling from our ADCIRC repo fork and in the future we may directly apply our patches to that branch

wwlwpd:
Issue 1705: listing v53release as what it really is, keeping existing v53release for now which
is identical except that it pulls from the upstream ADCIRC repo

Resolves #1705.

### DIFF
--- a/cloud/general/t/build-all-adcirc.sh
+++ b/cloud/general/t/build-all-adcirc.sh
@@ -13,6 +13,7 @@ if ! command -v init-adcirc.sh >/dev/null 2>&1; then
 fi
 
 versions=(
+  v53.05.live.0
   v53release
   v55.01-5bc04d6
   v55.02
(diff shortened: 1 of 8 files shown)


wwlwpd:
I am currently running through builds on supermikeiii - no changes to any patches were made, so as long as this builds and the version numbers/text are good then the PR is good

wwlwpd:
Oh wait because I updated the test script, this new "version" will get build automatically :-D

jasonfleming (approved):
Looks great thank you!

PR #1714, opened 2026-05-26
Upstream dir listing requires a trailing slash now.

### LINKED ISSUE #1713: GFS URL fix
NOAA updated their http configs, so now the directory listing requires a trailing forward slash.

wwlwpd:
Issue 1713: http_dir now uses the trailing slash to avoid a "forbidden" error, also fixed a "double slash" issue that was benign but looked bad.

Resolves #1713.

### DIFF
--- a/bin/get_gfs_status.pl
+++ b/bin/get_gfs_status.pl
@@ -55,7 +55,7 @@ GetOptions(
 
     sub http_ls {
         my $dir = shift;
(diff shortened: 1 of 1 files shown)


jasonfleming (approved):
I rebased this to the latest master and tested it successfully. Thank you @wwlwpd !

Issue #1731, opened 2026-07-06
Add data quality checks for upstream meteorological data feeds


jasonfleming:
We've been experiencing intermittent issues with data quality from upstream data sources, NAM and RRFS in particular. There have been empty grib2 files as well as grib2 files with empty `UGRD:10` records. Although it is difficult to troubleshoot intermittent issues effectively, additional data quality checks should be added that will hopefully catch these issues before they become show stoppers.

PR #1740, opened 2026-07-17
Adding trailing slash for directory listing.

### LINKED ISSUE #1738: Add required trailing slash for downloading NAM data
It appears that the NOMADS URL used when downloading NAM data has the same issue that was experienced with GFS: the URL for a directory listing requires a trailing slash (#1713). I have worked around this for now by setting the `BACKDIR` parameter in my ASGS config. The actual solution will be similar to the one for GFS.

wwlwpd:
Issue 1738: Upstream changes to the NOAA webserver configuration is refusing to list the file index without an explicit trailing slash. This commit adds that to the NAM dir listing request.

Resolves #1738.

### DIFF
--- a/bin/get_nam_status.pl
+++ b/bin/get_nam_status.pl
@@ -62,7 +62,7 @@ GetOptions(
 
     sub http_dir {
         my $dir = shift;
-        my $url         = sprintf( qq{https://nomads.ncep.noaa.gov%s}, $dir );
+        my $url         = sprintf( qq{https://nomads.ncep.noaa.gov%s/}, $dir );
         my $res         = $ua->get($url);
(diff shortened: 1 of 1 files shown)


jasonfleming (approved):
Looks awesome, thank you!

wwlwpd:
> Looks awesome, thank you!

I can merge it without testing, but I'd prefer verification that this does what was needed to resolve whoever was having issues. What do you think, @jasonfleming?

jasonfleming:
Yes I agree. I am also adding quality control checks on the grib2 files that we download (#1731) , so I am actively working on related things. I will test this and reply.

PR #1759, opened 2026-08-21
Alt-master including changes such as branching ensemble support, among others

jasonfleming:


wwlwpd (commented):
See review provided for in Discord.

writeProperties.sh
```diff
@@ -189,13 +189,25 @@ writeProperties()
    # runtime
    echo "path.rundir : $RUNDIR" >> $STORMDIR_RUN_PROPERTIES
    # each scenario
-   echo "path.fromdir : $FROMDIR" >> $STORMDIR_RUN_PROPERTIES
+   if [[ $SCENARIO != "branching"* ]]; then
+      echo "path.fromdir : $FROMDIR" >> $STORMDIR_RUN_PROPERTIES
+      echo "asgs.path.fromdir : $FROMDIR" >> $STORMDIR_RUN_PROPERTIES  # legacy property
+   else
+      branchName=${SCENARIO: -2}
+      if [[ $branchName == "03" && $branchName == "09" && $branchName == "15" ]]; then
```
wwlwpd, line 197: Impossible `03 && 09 && 15` condition

asgs_main.sh
```diff
@@ -1524,8 +1626,7 @@ submitJob()
          cp "$ADCIRCDIR/../adcirc.bin.buildinfo.json" . 2>> $SYSLOG
          echo "adcirc.file.metadata.build : adcirc.bin.buildinfo.json" >> run.properties
       fi
-      DATETIME=`date +'%Y-%h-%d-T%H:%M:%S'%z`
-      echo "time.${JOBTYPE}.start : $DATETIME" >> run.properties
+      echo "time.${JOBTYPE}.start : $(date +'%Y-%h-%d-T%H:%M:%S'%z)" >> run.properties
```
wwlwpd, line 1629: `$DATETIME` used without initialization (double check entire file)

asgs_main.sh
```diff
@@ -429,6 +485,16 @@ prep()
        fi
     fi
     if [[ $START == "hotstart" ]]; then
+       # save the value of FROMDIR if this is a branching
+       # ensemble (which will have different FROMDIRs depending
+       # on the branch)
+       nowcastFROMDIR=$FROMDIR
```
wwlwpd, line 491: Serial path unconditionally restores `FROMDIR=$nowcastFROMDIR` (double check entire file)

downloadHAFSB.sh
```diff
@@ -0,0 +1,420 @@
...
+fi
+
+DATE_URL="${BASE_URL}/hfsb.${DATE_UTC}"
+
+# ---------- Determine latest available cycle ----------
+
+if [[ -z "$CYCLE" ]]; then
+    echo "Checking HAFS-B cycles for ${DATE_UTC} ..."
+
+    PAGE="$(curl -fsSL --retry 3 --retry-delay 3 "${DATE_URL}/" || true)"
```
wwlwpd, line 249: Modern-curl assumption with `--retry-all-errors` (figure out a version safe method here, e.g., Frontera suffers from an "old" `curl`).

storm_track_gen.pl
```diff
@@ -104,9 +139,10 @@
            "year=s" => \$year,
            "coldstartdate=s" => \$coldstartdate,
            "hotstartseconds=s" => \$hotstartseconds,
+           "forecastend=s" => \$forecastend,
            "nws=s" => \$nws,
            "name=s" => \$name,
-           "method=s" => \$method,
```
wwlwpd, line 109: `--method` compatibility removed - double check this is updated anywhere that might have been using it

storm_track_gen.pl
```diff
@@ -64,6 +64,9 @@
 use Getopt::Long;
 use Date::Calc;
 use Cwd;
+use Geo::Ellipsoid;
+use Math::Trig;
```
wwlwpd, line 68: Did these get added to PERL-MODULES?

storm_track_gen.pl
```diff
@@ -297,6 +337,15 @@
       ASGSUtil::stderrMessage("INFO","The rMax variation is handled by the aswip program for the asymmetric models, and is therefore ignored by storm_track_gen.pl.",$test);
    }
 }
+if ( $name =~ /branching([0-9][0-9])/ ) {
```
wwlwpd, line 340: The code accepts any two-digit branch name and not just `01`-`17`

wwlwpd (changes requested):
Also, can you finish the `ENSTORM` → `SCENARIO` migration in the code paths touched here.

`SCENARIO` should be the variable used internally going forward. `ENSTORM` should only remain for backward compatibility with older configs and legacy output properties.

In practical terms:

* Read old `ENSTORM` values into `SCENARIO`.
* Use `SCENARIO` everywhere in the new and modified runtime logic.
* Keep `ENSTORM` only where older configs or legacy properties still require it.
* Make sure both old and new configs behave the same in tests.

This PR adds a lot of new scenario-specific behavior, so it is a good point to stop carrying both variables through the runtime logic.

The files this includes should just be,

* `asgs_main.sh`
* `writeProperties.sh`
* `bin/get-scenario-package-size`
* `t/config/test005_config.sh`

Attached is some [additional information that might be able to assist you](https://github.com/user-attachments/files/31714548/background.md).

writeProperties.sh
jasonfleming: Yes I'm not sure how this got in there, I will figure out what I was trying to do and fix it.

asgs_main.sh
jasonfleming: I don't see `$DATETIME` being used without initialization in this code snippet.

asgs_main.sh
jasonfleming: Not sure what is going on with this but I will track it down and fix it.

downloadHAFSB.sh
jasonfleming: I will fix the `curl` options so that we are more robust to different ambient `curl` versions.

storm_track_gen.pl
jasonfleming: Good catch. I have fixed this (only one shell script in `output`)

storm_track_gen.pl
jasonfleming: Yes these are in `PERL-MODULES`.

storm_track_gen.pl
jasonfleming: This is by design. The branching ensemble capability is still under active development. We may have several different ensemble designs and the acceptable ensemble member numbers will change based on the ensemble design.

asgs_main.sh
jasonfleming: Looking at the Discord reivew, I am probably going to just remove `$DATETIME` everywhere I see it (to the extent possible). It's better to just produce the date string inline.

jasonfleming:
Definitely agree on completing the migration from `ENSTORM` to `SCENARIO` in the code. I haven't done this up until now because it disturbs a lot of things. But I am more comfortable with this now. I don't think we need to deprecate old configuration files that use the `ENSTORM` parameter. I think we need to continue to accept them unchanged.

asgs_main.sh
wwlwpd: > Looking at the Discord reivew, I am probably going to just remove `$DATETIME` everywhere I see it (to the extent possible). It's better to just produce the date string inline.

I trust your eyes, don't make more work than necessary.

asgs_main.sh
wwlwpd: > Not sure what is going on with this but I will track it down and fix it.

If this doesn't make sense, you can defer.

downloadHAFSB.sh
wwlwpd: > I will fix the `curl` options so that we are more robust to different ambient `curl` versions.

I had to do this on Frontera to allow perl to continue to install, but just fix this particular instance don't go on a witch hunt.

PR #1761, opened 2026-08-31
Add fallback curl options for Frontera in `init-perlbew.sh`

wwlwpd:
@jasonfleming I ran into this oddly when rebuilding on Frontera. It's the first time I have seen it.

PR #1762, opened 2026-08-31
wsl-live updates

asgs-guest:
Various updates related to supporting WSL


jasonfleming:
This looks great so far. Not sure how I would test it for a review. Looking forward to using it.

wwlwpd:
I'll know if it's not working or broke something before I put it back into Ready for Review. Ultimately, the proof will be in the ADCIRC Live pudding.

wwlwpd:
I will re-confirm after we merge #1761.

